PR #1, opened 2017-04-12
removing cast prevents compile error

backesj:
I'm trying to build the cma34cr_ubuntu app, but I get a compile error unless I remove this cast.

### DIFF
--- a/apps/cma34cr_ubuntu/src/cross_vm_event.c
+++ b/apps/cma34cr_ubuntu/src/cross_vm_event.c
@@ -32,7 +32,7 @@ static camkes_emit_fn emitted_events[] = {
 };
 
 // mutex to protect shared event context
-static camkes_mutex_t cross_vm_event_mutex = (camkes_mutex_t) {
+static camkes_mutex_t cross_vm_event_mutex = {
     .lock = cross_vm_event_mutex_lock,
     .unlock = cross_vm_event_mutex_unlock,
 };


AdrianDanis:
Can you give any details on your system and what compiler version you are using?

backesj:
Hi Adrian,

Here is the compile error I'm getting (without removing the cast):

```
 [CC] cross_vm_dataport.o
 [CC] cross_vm_event.o
//home/john/pimcs/camkes-vm-manifest/apps/cma34cr_ubuntu/src/cross_vm_event.c:38:1: error: initializer element is not constant
 };
 ^
```

Here is my gcc version:

```
gcc (Ubuntu 4.9.4-2ubuntu1~14.04.1) 4.9.4
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

AdrianDanis:
Interesting that the complaint is "initializer element is not constant" when I feel it clearly is. We typically build with gcc >5 here, so that could be the difference. Either way I'll pull this in and merge it shortly

Smattr:
I believe the problem is that the right-hand side is a so-called compound literal (C99 §6.5.2.5). Apparently GCC allows these in initialisation [as an extension](https://gcc.gnu.org/onlinedocs/gcc/Compound-Literals.html). I don't know about this specific extension, but in my experience `-std=c99` in GCC means C99 plus the pieces of GNU99 the devs consider harmless. The definition of "harmless" seems to vary across GCC versions and vendor tweaks.

PR #3, opened 2017-08-15
Minor tweaks in Optiplex VM config

podhrmic:
* expose VESA frame buffer (Linux still doesn't know how to handle it, but it is the first step)
* disable hugepages so it can be run in Qemu


AdrianDanis:
While not particularly important your commit message implies that QEMU does not support huge page emulation, which is not true.

podhrmic:
Closing for timeout

PR #13, opened 2021-09-02
vmlinux.h: fix enum declaration

lsf37:
The previous declaration was declaring a type `IRQConstants` and (mistakenly) a variable `platform_interrupt_t`. Neither name is referenced in the rest of the code and `platform_interrupt_t` is the better name for the type.

### DIFF
--- a/components/VM_Arm/plat_include/tx2/plat/vmlinux.h
+++ b/components/VM_Arm/plat_include/tx2/plat/vmlinux.h
@@ -13,7 +13,7 @@
 #define GIC_LIC_INTID_BASE      (32)
 #define TX2_IRQ_PPI_VTIMER      (27)
 
-enum IRQConstants {
+typedef enum IRQConstants {
     TX2_TOP_TKE_SHARED0 = GIC_LIC_INTID_BASE,
     TX2_TOP_TKE_SHARED1,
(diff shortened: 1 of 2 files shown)


lsf37:
With this commit the VMs compile under gcc-10.

components/VM_Arm/plat_include/tx2/plat/vmlinux.h
```diff
@@ -13,7 +13,7 @@
 #define GIC_LIC_INTID_BASE      (32)
 #define TX2_IRQ_PPI_VTIMER      (27)
 
-enum IRQConstants {
+enum platform_interrupt_t {
```
kent-mcleod, line 16: This probably actually wants to be a typedef at the front, so that platform_interrupt_t can be used without the enum keyword. (https://docs.sel4.systems/processes/style-guide.html#type-conventions)

lsf37: Yes, that makes sense. Will fix.

lsf37: Done

kent-mcleod (approved):

PR #16, opened 2021-12-10
SMC Capability

Furao:
Add a new capability which certain threads can invoke so seL4 can make SMC calls on ARM platforms in EL2 (virtualized mode) on the thread’s behalf.

See https://sel4.atlassian.net/browse/RFC-9

This is dependent on these changes: https://github.com/seL4/seL4_projects_libs/pull/24 So it shouldn't be merged until that one is.


Furao:
@kent-mcleod This should be the last PR required for VMs to be able to have their SMC calls forwarded on. Just needs to be reviewed.

kent-mcleod (approved):

kent-mcleod:
Cool!

PR #18, opened 2022-02-03
libsel4vmmplatsupport: fix PCI passthrough on ARM

hlyytine:
libsel4vmmplatsupport: fix PCI passthrough on ARM

ARM uses the same PCI BAR emulation as x86 now.

Signed-off-by: Hannu Lyytinen <hannux@ssrc.tii.ae>

Test with: seL4/seL4_projects_libs#25

### DIFF
--- a/components/Init/src/virtio_net.c
+++ b/components/Init/src/virtio_net.c
@@ -127,8 +127,7 @@ void make_virtio_net(vm_t *vm, vmm_pci_space_t *pci, vmm_io_port_list_t *io_port
     emul_vm = vm;
 
     ioport_range_t virtio_port_range = {0, 0, VIRTIO_IOPORT_SIZE};
-    virtio_net = common_make_virtio_net(vm, pci, io_ports, virtio_port_range, IOPORT_FREE, 6, 6,
-                                        backend, true);
+    virtio_net = common_make_virtio_net(vm, pci, io_ports, virtio_port_range, IOPORT_FREE, 6, 6, backend);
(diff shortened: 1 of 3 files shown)


hlyytine:
See also https://github.com/seL4/seL4_projects_libs/pull/25

hlyytine:
Found out the ARM VM needs libvirtio linked in to access the platform specific headers. I still think getting the virtio interrupt pin and line from those headers is the best interim solution, hence updated commit. Seems that x86 build failed because it was being run without considering the changes in sel4_projects_libs.

kent-mcleod:
Thanks for making these changes!  I'll try and review them at some point over the next couple days.

hlyytine:
Still, it looks like the updated headers from my other PR are not used. I thought the Test-with added by axel-h would have taken care of that?

axel-h:
Test-with was using the URL and not the short form (but the GUI does not show this, see also https://github.com/seL4/ci-actions/issues/184), that is why this does not work. Should work now.

hlyytine:
After fixing a few more bugs in the vPCI code, turns out we can figure out the interrupt line and pin from the PCI config space itself, no need to clutter other code.

kent-mcleod (approved):
These changes are updating a breaking API change in https://github.com/seL4/seL4_projects_libs/pull/25.  It's probably worth allowing the breaking change because the argument being removed fixes an abstraction leak from an internal implementation constraint that's been resolved.

PR #20, opened 2022-03-09
VM_Arm: Use correct cell size

hlyytine:
Signed-off-by: Hannu Lyytinen <hannux@ssrc.tii.ae>


axel-h (approved):

axel-h:
Indeed, that seems a bug. I'm curious, did you find this from code review or was this causing invalid behavior somewhere?

hlyytine:
I don't remember all the details but IIRC we bumped onto some problems with RPi4 and autogenerated guest device trees. Not sure if the problem manifests with anything in the upstream (yet).

PR #22, opened 2022-03-28
init: merge two `guest_ram_regions` into one

abrandnewusername:
the upper bound of {0x500, 0x7c00 - 0x500} and the lower bound of
{0x7e00, 0x80000 - 0x7e00} belong to the same page. Merging them
to avoid vspace reservation failing.

### DIFF
--- a/components/Init/src/main.c
+++ b/components/Init/src/main.c
@@ -225,8 +225,10 @@ typedef struct memory_range {
 
 static memory_range_t guest_ram_regions[] = {
     /* Allocate all the standard low memory areas */
-    {0x500, 0x7c00 - 0x500},
-    {0x7e00, 0x80000 - 0x7e00},
+    /* On x86 the BIOS loads the MBR to 0x7c00. But for this VMM,
+     * we don't use MBR, so there is no need to exclude the MBR
+     * bootstrap code region */
+    {0x500, 0x80000 - 0x500},
     {0x80000, 0x9fc00 - 0x80000},
 };
 


components/Init/src/main.c
```diff
@@ -225,8 +225,11 @@ typedef struct memory_range {
 
 static memory_range_t guest_ram_regions[] = {
     /* Allocate all the standard low memory areas */
-    {0x500, 0x7c00 - 0x500},
-    {0x7e00, 0x80000 - 0x7e00},
+    /* {0x500, 0x7c00 - 0x500} and {0x7e00, 0x80000 - 0x7e00} are merged into one
+     * region because the upper bound of the first region and the lower bound of
+     * the second region are in the same page, so the vspace reservation will
+     * failed on the second region */
```
lsf37, line 231:
```suggestion
     * fail on the second region */
```

lsf37 (approved):
Is this related to the failure in [camkes-vm-examples](https://github.com/seL4/camkes-vm-examples/runs/5699666773?check_suite_focus=true)?

abrandnewusername:
> Is this related to the failure in [camkes-vm-examples](https://github.com/seL4/camkes-vm-examples/runs/5699666773?check_suite_focus=true)?

Yes. I corrected the return values in https://github.com/seL4/seL4_projects_libs/pull/38 so the vspace reservation for `guest_ram_regions` is finally able to fail, then I discovered the issue with `guest_ram_regions` itself.

Also, after fixing this, the size of `guest_ram_regions` changed, and I discovered https://github.com/seL4/seL4_projects_libs/pull/39.

It's fun that we keep uncovering bugs by fixing bugs :)

lsf37:
> It's fun that we keep uncovering bugs by fixing bugs :)

Many layers :-)

components/Init/src/main.c
```diff
@@ -225,8 +225,11 @@ typedef struct memory_range {
 
 static memory_range_t guest_ram_regions[] = {
     /* Allocate all the standard low memory areas */
-    {0x500, 0x7c00 - 0x500},
-    {0x7e00, 0x80000 - 0x7e00},
+    /* {0x500, 0x7c00 - 0x500} and {0x7e00, 0x80000 - 0x7e00} are merged into one
```
alisonfel, line 228: I might be lacking some historical context (our git history doesn't capture the introduction of these memory ranges). It seems `0x7C00`-`0x7DFF` was excluded to account for the MBR bootstrap code region. Though given we directly use Linux's 32-bit boot protocol (jumping directly into vmlinux entry point), I think the region should still be use-able memory in the end. 
As a minor nitpick, I think we could maybe extend the comment to name drop that we are now accounting for the MBR region, to give context as to why `{0x500, 0x7c00 - 0x500} and {0x7e00, 0x80000 - 0x7e00}` are merged into one.

axel-h: Yes, having some context here would be really helpful. Maybe even make a ASCII picture:
```
  --+--    +---------------------------+  0x00000     
    |      | 1280 bytes                |             
    |      +---------------------------+  0x00500     
    |      | 29184 byte                |            
    |      +---------------------------+  0x07700  (= 0x7c00 - 0x500)    
 512 KiB   | 1792 byte excluded        |
    |      | MBR is at 0x7C00-0x7E00   |
    |      +---------------------------+  0x07e00
    |      | 70400 byte                |             
    |      +---------------------------+  0x78200 (= 0x80000 - 0x7e00)
    |      | 32256 byte                |             
  --+--    +---------------------------+  0x80000
```

abrandnewusername:
Put some context in the commit message and comment.

components/Init/src/main.c
```diff
@@ -225,8 +225,9 @@ typedef struct memory_range {
 
 static memory_range_t guest_ram_regions[] = {
     /* Allocate all the standard low memory areas */
-    {0x500, 0x7c00 - 0x500},
-    {0x7e00, 0x80000 - 0x7e00},
+    /* The MBR is loaded at 0x7c00, but for this VMM, we don't use MBR
```
axel-h, line 228: Even is this example targets x86 anyway, I'd make this a bit more verbose: 
```suggestion
    /* On x86 the BIOS loads the MBR to 0x7c00. But for this VMM, we don't use
     * MBR, so there is no need to exclude the MBR bootstrap code region. */

alisonfel (approved):
LGTM :+1:

Issue #24, opened 2022-05-06
Do not run dispatch workflow action on github repo forks


axel-h:
Do not run dispatch workflow action on github repo forks, because they fail die to missing access rights
```
trigger.yaml
on push
Run seL4/ci-actions/trigger@master
/home/runner/work/_actions/seL[4](https://github.com/axel-h/camkes-vm/runs/6323832364?check_suite_focus=true#step:2:4)/ci-actions/master/js/../trigger/steps.sh
Need $INPUT_TOKEN for authentication.
Error: Action trigger failed.
```

This issue happen in other repos also, see e.g. https://github.com/seL4/docs/issues/135

PR #25, opened 2022-05-06
github: don't run dispatch on forks

axel-h:
Fixes https://github.com/seL4/camkes-vm/issues/24

Forks do not have a valid `secrets.PRIV_REPO_TOKEN`, thus the action will always fail.

### DIFF
--- a/.github/workflows/trigger.yml
+++ b/.github/workflows/trigger.yml
@@ -12,6 +12,7 @@ on:
 jobs:
   trigger:
     name: Repository Dispatch
+    if: ${{ github.repository_owner == 'seL4' }}
     runs-on: ubuntu-latest
     steps:
     - uses: seL4/ci-actions/trigger@master


lsf37:
You're right, sending dispatch events only makes sense from the seL4 org. However, that should be done inside the action itself, otherwise we have to update and maintain all repos.

It looks like we should move [this test here](https://github.com/seL4/ci-actions/blob/3d649f0e634593efb2ce3c652648fd6aeb18838d/trigger/dispatch.py#L23-L25) up and not fail, but succeed with a message that the dispatch event has not been generated.

axel-h:
The tricky thing is, that the check in `trigger/dispatch.py` wont run, even if moved up. Because the local action YAML contains this
```
      with:
        token: ${{ secrets.PRIV_REPO_TOKEN }}
```
which fails in forks becuause `secrets.PRIV_REPO_TOKEN` is missing. We'd have to pass a NULL value here then.

lsf37:
Need to double-check this, but I don't think the action fails if a secret is missing, it just gets an empty argument. In fact, the error message we're getting is the error message generated from the python script [here](https://github.com/seL4/ci-actions/blob/3d649f0e634593efb2ce3c652648fd6aeb18838d/trigger/dispatch.py#L15), so it seems the action started correctly, and then bailed out at this point because it explicitly checks for the secret.

axel-h:
Indeed, this seems to be from the script. However, if the script runs and returns an error detecting a forked repos, this still makes github send mail with the error. Thus, either this should not return an error, or the action itself should not run at all.

lsf37:
Yes, that's what I meant with shouldn't fail: it should return exit code 0, not 1, but should still print a message that it's skipping the dispatch.

lsf37:
Let me make the change there and then we can try another run in your fork.

axel-h:
Closing as this will not be fixed here.

PR #31, opened 2022-07-07
make DefineCAmkESVMFileServer() more flexible

axel-h:
- allow custom type and instance names
- allow passing files for `DefineCAmkESVMFileServer()`, this avoids calling `AddToFileServer()`
- `AddToFileServer()` becomes just a light wrapper
- use instance specific name for archive and lib

This allows doing customizing the file server component name and doing all in one call:
```
DefineCAmkESVMFileServer(
    INSTANCE "myFileServer"
    FILES
        "linux:${VM_IMG_LINUX}"
        "linux-initrd:${VM_IMG_ROOTFS}"
        "linux-dtb:${VM_IMG_DTB_PATCHED}"
    DEPENDS
        target_dtb_gen
)
```

See https://github.com/seL4/camkes-vm-examples/pull/37 for a usage examples.

### DIFF
--- a/camkes_vm_helpers.cmake
+++ b/camkes_vm_helpers.cmake
@@ -7,9 +7,6 @@
 cmake_minimum_required(VERSION 3.8.2)
 
 set(VM_PROJECT_DIR "${CMAKE_CURRENT_LIST_DIR}" CACHE INTERNAL "")
-if(NOT TARGET vm_fserver_config)
-    add_custom_target(vm_fserver_config)
-endif()
 
 # Function for declaring a CAmkESVM. This is called for each Init component in the applications
 # camkes config.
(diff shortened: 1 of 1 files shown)


wom-bat:
Need to investigate and fix the failing tests...

axel-h:
> Need to investigate and fix the failing tests...

Don't bother, I think I have a clue what fails there and what I have to improve to make it work. Just need to find some time to work in this again.

camkes_vm_helpers.cmake
```diff
@@ -63,18 +63,132 @@ function(DeclareCAmkESVM init_component)
...
+        else()
+            message(FATAL_ERROR "invalid format: ${item}")
+        endif()
+        # No need to pass dependencies here, we have already added them above.
+        AddToFileServer("${CPIO_NAME}" "${FILE_NAME}" INSTANCE "${PARAM_INSTANCE}")
+    endforeach()
+
+    # now process the file/deps list
+    get_target_property(files vm_fserver_config FILES_${PARAM_INSTANCE})
+    get_target_property(deps vm_fserver_config DEPS_${PARAM_INSTANCE})
```
axel-h, line 143: If the target property `DEPS_${PARAM_INSTANCE}` does not exist (because there are no dependencies), this will set `deps` to `deps-NOTFOUND` and then we see built errors lile `ninja: error: 'apps/Arm/vm_minimal/deps-NOTFOUND', needed by 'apps/Arm/vm_minimal/fserv/files/linux', missing and no known rule to make it`. Check for that case t avoid the error.

axel-h: Added a check for this cases and handle it. The case where we have no dependencies (and/or no files) works now as before.. There is no warning/error if there are no files, as this still creates a valid CPIO archive and it might be a valid use case in rare corner cases.

axel-h:
Could I get some feedback here, any chances to merge this?

kent-mcleod (approved):
Nice improvement! The ability to set CPIO output paths introduced in https://github.com/seL4/seL4_tools/pull/151/  will need to be merged before the second commit here can be merged.

axel-h:
Adding a feature that allows an empty file string also in `DefineCAmkESVMFileServer()`. This can be used for files that do no exist in every configuration. The caller can use a variable in this case instead of a fixed string, and leave the variable empty. This is used in the vm_minimal example, where not all platforms have initrd.

axel-h:
Avoid creating some unnecessary intermediate helper CMake targets.

axel-h:
Changed the order of the commits and split them up further. That allows cherry-picking the CPIO custom folder feature without taking the change in the CMake FileServer functions.

axel-h:
Removing the dependency of this PR on https://github.com/seL4/seL4_tools/pull/151 and seL4/camkes-vm-examples#37. It's not necessary to use a specific folder, if the instance name is simply used in the file name of the archive and lib. I will make another PR to move the build artifact out out of the build root folder, so this does not block this PR.

axel-h:
Can we merge this?

lsf37 (approved):
Since the dependency on https://github.com/seL4/seL4_tools/pull/151 is gone, I think we can merge this one, yes.

PR #36, opened 2022-09-13
virtio: console multiplex and vsock support

andybui01:
This modifies the virtio console to support multiplexing, as well as add a new virtio vsock device.

Test with: https://github.com/seL4/seL4_projects_libs/pull/88 and https://github.com/seL4/camkes-tool/pull/118

### DIFF
--- a/components/Init/src/main.c
+++ b/components/Init/src/main.c
@@ -54,6 +54,7 @@
 #include "virtio_net.h"
 #include "virtio_net_vswitch.h"
 #include "virtio_con.h"
+#include "virtio_vsock.h"
 
 #define BRK_VIRTUAL_SIZE 400000000
 #define ALLOCMAN_VIRTUAL_SIZE 400000000
--- a/components/Init/src/main.c
+++ b/components/Init/src/main.c
@@ -618,11 +618,6 @@ ioport_fault_result_t ioport_callback_handler(vm_vcpu_t *vcpu, unsigned int port
     return result;
 }
 
-void make_virtio_net_vswitch(vm_t *vm)
-{
-    return make_virtio_net_vswitch_driver(vm, pci, io_ports);
-}
-
 void *main_continued(void *arg)
 {
     int error;
--- a/components/Init/src/main.c
+++ b/components/Init/src/main.c
@@ -53,6 +53,7 @@
 #include "fsclient.h"
 #include "virtio_net.h"
 #include "virtio_net_vswitch.h"
+#include "virtio_con.h"
 
 #define BRK_VIRTUAL_SIZE 400000000
 #define ALLOCMAN_VIRTUAL_SIZE 400000000
@@ -586,11 +587,16 @@ void init_con_irq_init(void)
     device_notify_list_len = irqs;
     device_notify_list = malloc(sizeof(*device_notify_list) * irqs);
     ZF_LOGF_IF(device_notify_list == NULL, "Malloc failed");
-    for (i = 0; i < irqs; i++) {
-        init_cons_has_interrupt(i, &badge, &fun);
-        device_notify_list[i].badge = badge;
-        device_notify_list[i].func = (void (*)(vm_t *))fun;
+
+    int notify_idx = 0;
+    for (i = 0; i < init_cons_num_connections(); i++) {
+        if (init_cons_has_interrupt(i, &badge, &fun)) {
+            device_notify_list[notify_idx].badge = badge;
+            device_notify_list[notify_idx].func = (void (*)(vm_t *))fun;
+            notify_idx++;
+        }
     }
+    assert(notify_idx == irqs);
 }
 
 ioport_fault_result_t ioport_callback_handler(vm_vcpu_t *vcpu, unsigned int port_no, bool is_in, unsigned int *value,
--- a/components/Init/src/virtio_blk.c
+++ b/components/Init/src/virtio_blk.c
@@ -33,6 +33,7 @@
 
 #include "vm.h"
 #include "virtio_blk.h"
+#include "virtio_irq.h"
 
 #define VIRTIO_VENDOR_ID            0x1af4
 #define VIRTIO_DEVICE_ID            0x1001
@@ -40,7 +41,6 @@
 #define VIRTIO_BLK_IOBASE           0x8000
 #define VIRTIO_QUEUE_SIZE           128
 #define VIRTIO_BLK_DISK_BLK_SIZE    512
-#define VIRTIO_BLK_IRQ              7
 #define VIRTIO_BLK_SIZE_MAX         4096
 #define VIRTIO_BLK_SEG_MAX          1
 
--- a/components/Init/src/virtio_con.c
+++ b/components/Init/src/virtio_con.c
@@ -46,6 +46,7 @@
 #include <camkes/virtqueue.h>
 
 #include "virtio_con.h"
+#include "virtio_irq.h"
 
 static virtio_con_t *virtio_con;
 
@@ -198,7 +199,7 @@ static void console_handle_irq(void *cookie)
         return;
     }
 
-    int err = vm_inject_irq(virtio_cookie->vm->vcpus[BOOT_VCPU], 9);
+    int err = vm_inject_irq(virtio_cookie->vm->vcpus[BOOT_VCPU], VIRTIO_CON_IRQ);
     if (err) {
         ZF_LOGE("Failed to inject irq");
         return;
@@ -221,7 +222,7 @@ void make_virtio_con(vm_t *vm, vmm_pci_space_t *pci, vmm_io_port_list_t *io_port
     backend.console_data = (void *)console_cookie;
     ioport_range_t virtio_port_range = {0, 0, VIRTIO_IOPORT_SIZE};
     virtio_con = common_make_virtio_con(vm, pci, io_ports, virtio_port_range, IOPORT_FREE,
-                                        9, 9, backend);
+                                        VIRTIO_CON_IRQ, VIRTIO_CON_IRQ, backend);
     console_cookie->virtio_con = virtio_con;
     console_cookie->vm = vm;
 
--- a/components/Init/src/virtio_con.c
+++ b/components/Init/src/virtio_con.c
@@ -1,11 +1,11 @@
 /*
- * Copyright 2019, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
 /**
- * @brief virtio console backend layer for ARM
+ * @brief virtio console backend layer for x86
  *
  * The virtio console backend has two layers, the emul layer and the backend layer.
  * This file is part of the backend layer, it's responsible for:
@@ -21,14 +21,13 @@
  * being configured correctly in camkes.
  *
  * @todo Handling the RX and TX is similar between the x86/ARM VMMs, hence this file
- * and `components/Init/src/virtio_con.c` share a lot of common code. A refactor for
- * the backend layer is needed.
+ * and `components/VM_Arm/src/modules/virtio_con.c` share a lot of common code.
+ * A refactor for the backend layer is needed.
  */
 
 #include <stdint.h>
 #include <string.h>
 #include <autoconf.h>
-#include <vmlinux.h>
 
 #include <utils/util.h>
 
@@ -37,20 +36,27 @@
 
 #include <sel4vmmplatsupport/drivers/virtio_con.h>
 #include <sel4vmmplatsupport/device.h>
-#include <sel4vmmplatsupport/arch/vpci.h>
+
+#include <sel4vm/guest_irq_controller.h>
+#include <sel4vm/boot.h>
 
 #include <platsupport/serial.h>
-#include <virtioarm/virtio_console.h>
 
 #include <virtqueue.h>
 #include <camkes/virtqueue.h>
 
-/* Number of ports is the number of crossvm connections plus hvc0 */
-#define NUM_PORTS (ARRAY_SIZE(serial_layout) + 1)
+#include "virtio_con.h"
 
-static virtio_con_t *virtio_con = NULL;
-extern vmm_pci_space_t *pci;
-extern vmm_io_port_list_t *io_ports;
+static virtio_con_t *virtio_con;
+
+/* Used as the parameter for callback functions (currently only `console_handle_irq`) */
+typedef struct virtio_con_cookie {
+    virtio_con_t *virtio_con;
+    vm_t *vm;
+} virtio_con_cookie_t;
+
+/* Configured in camkes */
+#define NUM_PORTS ARRAY_SIZE(serial_layout)
 
 /* 4088 because the serial_shmem_t struct has to be 0x1000 bytes big */
 #define BUFSIZE (0x1000 - 2 * sizeof(uint32_t))
@@ -85,7 +91,7 @@ typedef struct serial_shmem {
 } serial_shmem_t;
 compile_time_assert(serial_shmem_4k_size, sizeof(serial_shmem_t) == 0x1000);
 
-/*
+/**
  * Represents the virtqueues used in a given link between
  * two VMs.
  */
@@ -114,21 +120,17 @@ static void virtio_con_notify_recv(int port)
 /**
  * Callback function for camkes virtqueue notification. Polls each receive ring buffer
  * to see if there are data pending. For some legacy reasons, this callback passes a
- * VMM handler and a cookie as parameters, they're unused here.
+ * VMM handler as a parameter, it's unused here.
  *
- * @see include/vmlinux.h (async_event_handler_fn_t) for usages
- *
- * @return Returns 0 on success.
+ * @see main.c (struct device_notify) for usages
  */
-static int handle_serial_console(vm_t *vmm, void *cookie UNUSED)
+void handle_serial_console(vm_t *vmm)
 {
-    /* Poll each ring buffer to see if data was added */
     for (int i = 0; i < NUM_PORTS; i++) {
         if (connections[i].recv_buf->head != connections[i].recv_buf->tail) {
             virtio_con_notify_recv(i);
         }
     }
-    return 0;
 }
 
 /**
@@ -143,77 +145,91 @@ static void tx_virtcon_forward(int port, char c)
         return;
     }
 
-    serial_shmem_t *buffer = connections[port].send_buf;
-    buffer->buf[buffer->tail] = c;
+    serial_shmem_t *batch_buf = connections[port].send_buf;
+    batch_buf->buf[batch_buf->tail] = c;
     __atomic_thread_fence(__ATOMIC_ACQ_REL);
-    buffer->tail = (buffer->tail + 1) % BUFSIZE;
+    batch_buf->tail = (batch_buf->tail + 1) % BUFSIZE;
 
     /* If newline or staging area full, it's time to send */
-    if (c == '\n' || (buffer->tail + 1) % BUFSIZE == buffer->head) {
+    if (c == '\n' || (batch_buf->tail + 1) % BUFSIZE == batch_buf->head) {
         __atomic_thread_fence(__ATOMIC_ACQ_REL);
         connections[port].notify();
     }
 }
 
-/* camkes-generated symbols */
-/* regions shared by the SerialServer and the VMM */
-extern serial_shmem_t *batch_buf;
-extern serial_shmem_t *serial_getchar_buf;
-
-extern seL4_Word serial_getchar_notification_badge(void);
-
 /**
- * Initialize a virtio console backend. This includes:
- * - adding an IO port handler
- * - adding a PCI entry
- * - initializing the emul layer
- * - setting up connections (includes hvc0)
- *
- * @see include/vmlinux.h (vmm_module_t) for usages
- *
- * @param vm The vm handler of the caller
- * @param cookie Unused in this function
+ * Sets up virtio console connections.
  */
-void make_virtio_con(vm_t *vm, void *cookie UNUSED)
+static void make_virtio_con_virtqueues(void)
 {
-    ZF_LOGF_IF((NUM_PORTS > VIRTIO_CON_MAX_PORTS), "Too many ports configured (up the constant)");
-
-    virtio_con = virtio_console_init(vm, tx_virtcon_forward, pci, io_ports);
-    ZF_LOGF_IF(!virtio_con, "Failed to initialise virtio console");
-
     int err;
     for (int i = 0; i < NUM_PORTS; i++) {
-        if (i == 0) {
-            /* Port 0 is the existing SerialServer shmem interface */
-            connections[0].notify = batch_batch;
-            connections[0].recv_buf = serial_getchar_buf;
-            connections[0].send_buf = batch_buf;
-            err = register_async_event_handler(serial_getchar_notification_badge(), handle_serial_console, NULL);
-            if (err) {
-                ZF_LOGE("Failed to register event handler");
-                return;
-            }
-            continue;
-        }
-
-        camkes_virtqueue_channel_t *recv_channel = get_virtqueue_channel(VIRTQUEUE_DEVICE, serial_layout[i - 1].recv_id);
-        camkes_virtqueue_channel_t *send_channel = get_virtqueue_channel(VIRTQUEUE_DRIVER, serial_layout[i - 1].send_id);
+        camkes_virtqueue_channel_t *recv_channel = get_virtqueue_channel(VIRTQUEUE_DEVICE, serial_layout[i].recv_id);
+        camkes_virtqueue_channel_t *send_channel = get_virtqueue_channel(VIRTQUEUE_DRIVER, serial_layout[i].send_id);
 
         if (!recv_channel || !send_channel) {
             ZF_LOGE("Failed to get channel");
             return;
         }
 
-        err = register_async_event_handler(recv_channel->recv_badge, handle_serial_console, NULL);
-        if (err) {
-            ZF_LOGE("Failed to register event handler");
-            return;
-        }
-
         connections[i].notify = send_channel->notify;
         connections[i].recv_buf = (serial_shmem_t *) recv_channel->channel_buffer;
         connections[i].send_buf = (serial_shmem_t *) send_channel->channel_buffer;
+        connections[i].recv_buf->head = 0;
+        connections[i].recv_buf->tail = 0;
+        connections[i].send_buf->head = 0;
+        connections[i].send_buf->tail = 0;
     }
 }
 
-DEFINE_MODULE(virtio_con, NULL, make_virtio_con)
+/**
+ * Callback function for the emul layer. Notifies the guest that there is
+ * something in its used ring
+ *
+ * @see virtio_pci_console.h for function format
+ *
+ * @param cookie Pointer to a virtio_con_cookie_t
+ */
+static void console_handle_irq(void *cookie)
+{
+    virtio_con_cookie_t *virtio_cookie = (virtio_con_cookie_t *)cookie;
+    if (!virtio_cookie || !virtio_cookie->vm) {
+        ZF_LOGE("NULL virtio cookie given to raw irq handler");
+        return;
+    }
+
+    int err = vm_inject_irq(virtio_cookie->vm->vcpus[BOOT_VCPU], 9);
+    if (err) {
+        ZF_LOGE("Failed to inject irq");
+        return;
+    }
+}
+
+void make_virtio_con(vm_t *vm, vmm_pci_space_t *pci, vmm_io_port_list_t *io_ports)
+{
+    ZF_LOGF_IF((NUM_PORTS > VIRTIO_CON_MAX_PORTS), "Too many ports configured (up the constant)");
+
+    int err;
+    struct virtio_console_passthrough backend;
+
+    backend.handleIRQ = console_handle_irq;
+    backend.backend_putchar = tx_virtcon_forward;
+
+    virtio_con_cookie_t *console_cookie = (virtio_con_cookie_t *)calloc(1, sizeof(virtio_con_cookie_t));
+    ZF_LOGF_IF(console_cookie == NULL, "Failed to allocated virtio console cookie");
+
+    backend.console_data = (void *)console_cookie;
+    ioport_range_t virtio_port_range = {0, 0, VIRTIO_IOPORT_SIZE};
+    virtio_con = common_make_virtio_con(vm, pci, io_ports, virtio_port_range, IOPORT_FREE,
+                                        9, 9, backend);
+    console_cookie->virtio_con = virtio_con;
+    console_cookie->vm = vm;
+
+    make_virtio_con_virtqueues();
+    assert(virtio_con);
+}
+
+/**
+ * Dummy function used to register irq callback handlers.
+ */
+void make_virtio_con_driver_dummy(vm_t *vm, vmm_pci_space_t *pci, vmm_io_port_list_t *io_ports) {}
--- a/components/Init/src/virtio_irq.h
+++ b/components/Init/src/virtio_irq.h
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+/**
+ * Hardcoded IRQ numbers for virtual devices. These have to be unique
+ * for each device.
+ * Where Linux has a virtual IRQ number assigned (in `/proc/interrupts`)
+ * we use that, otherwise pick a free one.
+ *
+ * @todo: Having hardcoded numbers for virtio devices might become
+ * an issue as the list of virtual devices grows. Might be worth it
+ * to have an IRQ number allocator.
+ */
+#define TIMER_IRQ                   0
+/* #define CASCADE_IRQ                 2 */
+#define TTYS0_IRQ                   4
+#define VIRTIO_NET_IRQ              6
+#define VIRTIO_BLK_IRQ              7
+#define RTC_IRQ                     8
+#define VIRTIO_CON_IRQ              9
+#define VIRTIO_SCK_IRQ              10
--- a/components/Init/src/virtio_vsock.c
+++ b/components/Init/src/virtio_vsock.c
@@ -1,0 +1,255 @@
+/*
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/**
+ * @brief virtio vsock backend layer for x86
+ *
+ * The virtio vsock backend has two layers, the emul layer and the backend layer.
+ * This file is part of the backend layer, it's responsible for:
+ *
+ * - TX: receiving data from the emul layer and forwarding them to the destination
+ *   VM using camkes virtqueue (which is different from the virtqueue that's
+ *   used to communicate between the guest and the emul layer)
+ * - RX: receiving data from the source port via camkes virtqueue and invoking the
+ *   emul layer to handle the data
+ * - virtio vsock backend initialization
+ * - sending virtual IRQ to the guest
+ *
+ * Data flow:
+ *
+ *          vq                      camkes vq                      vq
+ *  src VM <--> (emul <-> backend) <---------> (backend <-> emul) <--> dest VM
+ *                  src VMM                          dest VMM
+ *
+ * This file relies on the connection topology and camkes virtqueue connection
+ * being configured correctly in camkes.
+ */
+
+#include <stdint.h>
+#include <string.h>
+#include <autoconf.h>
+
+#include <utils/util.h>
+
+#include <camkes.h>
+#include <camkes/dataport.h>
+
+#include <sel4vm/guest_irq_controller.h>
+#include <sel4vm/boot.h>
+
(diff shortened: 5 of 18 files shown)


wom-bat:
depends on https://github.com/seL4/util_libs/pull/136 https://github.com/seL4/seL4_projects_libs/pull/74 and https://github.com/seL4/global-components/pull/39

components/Init/src/main.c
```diff
@@ -615,6 +598,16 @@ void make_virtio_net_vswitch(vm_t *vm)
     return make_virtio_net_vswitch_driver(vm, pci, io_ports);
 }
 
+void make_virtio_con(vm_t *vm)
+{
+    return make_virtio_con_driver(vm, pci, io_ports);
```
wom-bat, line 603: Why is there this indirection here?

components/Init/src/virtio_blk.c
```diff
@@ -17,10 +17,10 @@
 #include <camkes.h>
 #include <camkes/dataport.h>
 
-#include <ethdrivers/virtio/virtio_pci.h>
-#include <ethdrivers/virtio/virtio_net.h>
-#include <ethdrivers/virtio/virtio_ring.h>
-#include <satadrivers/virtio/virtio_blk.h>
+#include <virtio/virtio_pci.h>
```
wom-bat, line 20: Ideally this refactoring would be a separate PR.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
+/*
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
```
wom-bat, line 6: Please add a comment to make it obvious what this file is.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+typedef struct virtio_con_cookie {
+    virtio_con_t *virtio_con;
+    vm_t *vm;
+} virtio_con_cookie_t;
+
+/* Configured in camkes */
+#define NUM_PORTS ARRAY_SIZE(serial_layout)
+
+/* 4088 because the serial_shmem_t struct has to be 0x1000 bytes big */
+#define BUFSIZE 4088
```
wom-bat, line 42: Is it 4096 bytes or one page? If it's one page, then use the PAGESIZE macro (whatever it's called)

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+ */
+typedef struct serial_conn {
+    void (*notify)(void);
+    volatile serial_shmem_t *recv_buf;
+    volatile serial_shmem_t *send_buf;
+} serial_conn_t;
+static serial_conn_t connections[NUM_PORTS];
+
+/* This is a temporary buffer used to stage data before completing the virtio RX */
+char temp_rx_buf[BUFSIZE];
```
wom-bat, line 64: declare `static` if it's not visible outside this file; name something less likely to collide if it is meant to be visible.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+
+void virtio_console_putchar(int port, virtio_emul_t *con, char *buf, int len) WEAK;
+static int virtio_con_notify_recv(int port)
+{
+    volatile serial_shmem_t *buffer = connections[port].recv_buf;
+    int count = 0;
+    while (buffer->head != buffer->tail) {
+        temp_rx_buf[count] = buffer->buf[buffer->head];
```
wom-bat, line 72: Is it possible to avoid this copy?

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+static void tx_virtcon_forward(int port, char c)
+{
+#if CONFIG_NUM_DOMAINS > 1 && CONFIG_PRINTING
+    seL4_DebugPutChar(c);
+#else
+    if (port >= NUM_PORTS) {
+        return;
+    }
+
+    volatile serial_shmem_t *batch_buf = connections[port].send_buf;
```
wom-bat, line 100: Don't think this needs to be volatile.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+    for (int i = 0; i < NUM_PORTS; i++) {
+        if (connections[i].recv_buf->head != connections[i].recv_buf->tail) {
+            virtio_con_notify_recv(i);
+        }
+    }
+}
+
+static void tx_virtcon_forward(int port, char c)
+{
+#if CONFIG_NUM_DOMAINS > 1 && CONFIG_PRINTING
```
wom-bat, line 93: Not sure what this conditional is doing here.  It seems to bypass the function if you can print via the kernel, and have more than one VMM?

components/Init/src/virtio_irq.h
```diff
@@ -0,0 +1,16 @@
+/*
```
wom-bat, line 1: this cleanup and in the next file should be a separate PR.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,152 @@
+/*
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#include <stdio.h>
```
wom-bat, line 6: Do you actually need stdio.h?

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,152 @@
...
+
+/* 4092 because the vsock_shmem_t struct has to be 0x1000 bytes big */
+#define BUFSIZE 4092
+
+/* This struct occupies exactly 1 page and represents the data in the shmem region */
+typedef struct vsock_shmem {
+    uint32_t len;
+    char buf[BUFSIZE];
+} vsock_shmem_t;
+compile_time_assert(vsock_shmem_correct_size, sizeof(vsock_shmem_t) == 0x1000);
```
wom-bat, line 45: Why not #define BUFSIZE (PAGE_SIZE - sizeof (uint32_t_)) ?

components/Init/src/virtio_vsock.h
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <sel4vmmplatsupport/ioports.h>
+#include <sel4vmmplatsupport/drivers/pci.h>
+
+#define VIRTIO_IOPORT_SIZE  0x400
```
wom-bat, line 12: Please add comment as to what this is.

components/VM_Arm/src/modules/virtio_con.c
```diff
@@ -22,61 +22,117 @@
...
+static int virtio_con_notify_recv(int port, virtqueue_device_t *queue)
+{
+    volatile serial_shmem_t *buffer = connections[port].recv_buf;
     int count = 0;
     while (buffer->head != buffer->tail) {
-        txbuf[count % BUFSIZE] = buffer->buf[buffer->head];
+        temp_rx_buf[count] = buffer->buf[buffer->head];
         buffer->head = (buffer->head + 1) % BUFSIZE;
-        count += 1 % BUFSIZE;
+        count++;
```
wom-bat, line 67: This could overrun the buffer as txbuf[count] chases head.  It should also be possible to rework this to be multicore safe.
And again, do we absolutely need to do an extra copy here?

wom-bat (commented):
General comment: I'd like to improve the documentation of all this stuff, buy adding a comment for each type, global variable, and non-trivial function, as to what it is for, and docs for function arguments and return values.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,152 @@
...
+            vsock_rx_complete(virtio_vsock->emul, (char *) connections[i].recv_buf->buf, connections[i].recv_buf->len);
+            connections[i].recv_buf->len = 0;
+        }
+    }
+}
+
+static void tx_forward(int cid, void *buffer, unsigned int len)
+{
+    /* Find the CID and port */
+    for (int i = 0; i < NUM_PORTS; i++) {
```
abrandnewusername, line 74: Is it possible to avoid doing a for loop every time?

components/Sataserver/include/plat/pc99/supermicro/plat/sata.h
```diff
@@ -0,0 +1,23 @@
+/*
```
abrandnewusername, line 1: I think this commit should be in a separated pull request.

abrandnewusername (commented):
Seems like there are a lot of common code between [components/VM_Arm/src/modules/virtio_con.c](https://github.com/seL4/camkes-vm/pull/36/files#diff-3265d75a21c89137c4647bd28d5d75b4db0b59e7e87a986636a1130557d545ea) and [components/Init/src/virtio_con.c](https://github.com/seL4/camkes-vm/pull/36/files#diff-98b5c9eefaed33c6cd04830d9d0157ce0ccd3c16fb0ad6ee5c1fa4782316ddf2), is it really necessary?

components/Init/src/virtio_irq.h
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#define VIRTIO_NET_IRQ              6
```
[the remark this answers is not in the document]
andybui01: Nowhere in particular, we just have to pick a unique number for the virtual IRQ.

components/Sataserver/include/plat/pc99/supermicro/plat/sata.h
andybui01: It was, it's just polluted my branch, oops.

components/Init/src/virtio_con.c
andybui01: It's 4096 bytes. CAmkES demands shmem buffers be a multiple of 0x1000 bytes.

components/Init/src/virtio_con.c
andybui01: > Seems like there are a lot of common code between [components/VM_Arm/src/modules/virtio_con.c](https://github.com/seL4/camkes-vm/pull/36/files#diff-3265d75a21c89137c4647bd28d5d75b4db0b59e7e87a986636a1130557d545ea) and [components/Init/src/virtio_con.c](https://github.com/seL4/camkes-vm/pull/36/files#diff-98b5c9eefaed33c6cd04830d9d0157ce0ccd3c16fb0ad6ee5c1fa4782316ddf2), is it really necessary?

I've just been following the set up for the virtio backends in the VMM so far. These files need a refactor (in addition with virtio net). Handling the RX and TX is similar between the x86/ARM VMMs, and the majority of the difference comes from setting up the virtual IRQs for the guest OS.

components/VM_Arm/src/modules/virtio_con.c
andybui01: temp_rx_buf is BUFSIZE bytes big. the shmem buffer is also BUFSIZE bytes big, so as long as both buffers are the same size, we can just check that the shmem head is not at the tail.

The extra copy can be avoided if the virtio emul layer knows that the shmem area is a circular buffer.

components/Init/src/main.c
andybui01: All the make_virtio_X functions exist in main.c currently. They could be defined within their own respective files, however.

andybui01:
As per the comments on https://github.com/seL4/global-components/pull/39, I've rejigged the code to provide backwards compatibility for the existing SerialServer on ttyS0 (x86) and hvc0 (Arm).

andybui01:
@wom-bat are you happy with the changes right now? A refactor effort can be done in a separate pull request to reduce arch-differences in the VMM at a later date.

components/Init/src/virtio_con.c
wom-bat: Please add a comment to say what this file is for!

components/Init/src/virtio_con.c
wom-bat: It should be possible not to have this copy here.

components/Init/src/virtio_con.c
wom-bat: Please add a comment to explain what;s going on.

components/Init/src/virtio_irq.h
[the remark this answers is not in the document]
wom-bat: Add a comment to say so.  Particularly if there are cross-dependencies ... (can't reuse the same IRQ between components for instance).  Where are the others defined?  How can you be sure someone won't define something else as IRQ 8?

components/VM_Arm/src/modules/virtio_con.c
wom-bat: Make it so ... make the virtIOemul layers understand the shape of the buffers it is dealing with.

components/VM_Arm/src/modules/virtio_con.c
```diff
@@ -20,63 +19,133 @@
...
+    if (c == '\n' || (batch_buf->tail + 1) % BUFSIZE == batch_buf->head) {
+        connections[port].notify();
     }
 #endif
 }
 
+/* regions shared by the SerialServer and the VMM */
+extern serial_shmem_t volatile *batch_buf;
+extern serial_shmem_t volatile *serial_getchar_buf;
+
```
wom-bat, line 108: It's confusing to have the same-named extern variables here and local in tx_virtcon_forward()

wom-bat (changes requested):

components/Init/src/main.c
wom-bat: Why do we have both `make_virtio_con()` and `make_virtio_con_driver()`?
Why not just rename `make_virtio_con_driver()` as `make_virtio_con()` ????

components/Init/src/virtio_con.c
wom-bat: Please a add a big comment at the start of this file to say what this file is all about.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+
+#include "virtio_con.h"
+#include "virtio_irq.h"
+
+/*
+ * This file is part of the virtio console backend, it's responsible for
+ * sending/receiving data to/from other VM's through using our virtqueue
+ * implementation.
+ */
+
```
wom-bat, line 35: You could move this comment to the top of the file as part of the BIG COMMENT i've been asking for.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+static virtio_con_t *virtio_con;
+
+typedef struct virtio_con_cookie {
+    virtio_con_t *virtio_con;
+    vm_t *vm;
+} virtio_con_cookie_t;
+
```
wom-bat, line 42: What is this datastructure for?  Please add a comment.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,175 @@
...
+/* 4088 because the serial_shmem_t struct has to be 0x1000 bytes big */
+#define BUFSIZE (0x1000 - 2 * sizeof(uint32_t))
+
+/* This struct occupies exactly 1 page and represents the data in the shmem region */
+typedef struct serial_shmem {
+    uint32_t head;
+    uint32_t tail;
+    char buf[BUFSIZE];
+} serial_shmem_t;
```
wom-bat, line 54: Please expand the comment to say what head and tail do, and that access can be lock free (in what circumstances?) and that head and tail increase continually, are used mod BUFSIZE to index into buf

components/Init/src/virtio_irq.h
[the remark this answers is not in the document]
wom-bat: This *really* needs that comment.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#include <stdint.h>
```
wom-bat, line 6: Add a honking-big comment here to say what this file is for, please.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,183 @@
...
+ * Represents the virtqueues used in a given link between
+ * two VMs.
+ */
+typedef struct conn {
+    int cid;
+    virtqueue_driver_t *send_queue;
+    virtqueue_device_t *recv_queue;
+} conn_t;
+static conn_t connections[NUM_PORTS];
+
```
wom-bat, line 48: General comment: I'd like to see doxygen style comments on *every* function that say what the function is for, what its arguments are, and any pre/post conditions.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,183 @@
...
+    virtqueue_ring_object_t handle;
+    while (virtqueue_get_used_buf(queue, &handle, &wr_len)) {
+        while (camkes_virtqueue_driver_gather_buffer(queue, &handle, &buf, &buf_size, &flag) >= 0) {
+            /* Clean up and free the buffer we allocated */
+            camkes_virtqueue_buffer_free(queue, buf);
+        }
+    }
+}
+
+void vsock_rx_complete(virtio_emul_t *emul, char *buf, unsigned len) WEAK;
```
wom-bat, line 63: Is here only one completion call per system, or can you have one per vsock?   Plus AFAIK  `WEAK` is only effective on a sumbol definition, not a declaration.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,183 @@
...
+
+static void vsock_inject_irq(void *cookie)
+{
+    virtio_vsock_cookie_t *virtio_cookie = (virtio_vsock_cookie_t *)cookie;
+    if (!virtio_cookie) {
+        ZF_LOGE("NULL virtio cookie given to raw irq handler");
+        return;
+    }
+
+    int err = vm_inject_irq(virtio_cookie->vm->vcpus[BOOT_VCPU], VIRTIO_SCK_IRQ);
```
wom-bat, line 148: We *always* inject into the BOOT VCPU, instead of allowing the guest to choose a CPU to be interrupted via emulated the IOAPIC/GIC ?  Why?

components/VM_Arm/src/modules/virtio_con.c
wom-bat: extern variables ideally are in header files.

wom-bat (changes requested):
Please comment to a higher standard: the customer wants Doxygen style comments everywhere.

components/Init/src/main.c
andybui01: I believe the original intent was to have all CAmkES interacting functions within `camkes-vm`, so all functions called by CAmkES are defined and implemented within `camkes-vm`.

components/Init/src/virtio_irq.h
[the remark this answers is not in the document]
andybui01: Never got around to this but originally IRQ numbers were sparsely designated and this commit aggregated all the magic values into a single header file. A better approach would be to have an irq number allocator.

components/VM_Arm/src/modules/virtio_con.c
abrandnewusername: It seems to be a "conventional" way to deal with camkes-generated symbols (putting them right above where they're needed).

axel-h: Actually not, the proper header templates seem to be missing sometimes.

components/Init/src/virtio_vsock.c
abrandnewusername: I'm not sure if the current VMM code handles IRQ affinity changing for the guest, if it does, can anyone point me to the right place in the VMM code?

abrandnewusername:
Added comments and some minor changes. @wom-bat can you please (re-)review this?

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,227 @@
...
+
+ * Way of indexing: `head` and `tail` should not increase beyond (BUFSIZE - 1), as
+ * they represent the real index in `buf`.
+ */
+typedef struct serial_shmem {
+    uint32_t head;
+    uint32_t tail;
+    char buf[BUFSIZE];
+} serial_shmem_t;
+compile_time_assert(serial_shmem_1k_size, sizeof(serial_shmem_t) == 0x1000);
```
axel-h, line 82:
```suggestion
compile_time_assert(serial_shmem_4k_size, sizeof(serial_shmem_t) == 0x1000);
```

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,227 @@
...
+    }
+}
+
+/**
+ * Callback function for the emul layer. Notifies the guest that there is
+ * something in its used ring
+ *
+ * @see virtio_pci_console.h for function format
+ *
+ * @param cookie Assumes struct virtio_con_cookie
```
axel-h, line 179:
```suggestion
 * @param cookie pointer to a virtio_con_cookie_t
```

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,227 @@
...
+ */
+static void console_handle_irq(void *cookie)
+{
+    virtio_con_cookie_t *virtio_cookie = (virtio_con_cookie_t *)cookie;
+    if (!virtio_cookie) {
```
axel-h, line 184:
```suggestion
    if (!virtio_cookie || !virtio_cookie->vm) {
```

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,243 @@
...
+    virtqueue_ring_object_t handle;
+
+    while (virtqueue_get_used_buf(queue, &handle, &wr_len)) {
+        while (camkes_virtqueue_driver_gather_buffer(queue, &handle, &buf, &buf_size, &flag) >= 0) {
+            camkes_virtqueue_buffer_free(queue, buf);
+        }
+    }
+}
+
+/* Should be defined in virtio_vsock_emul.c */
```
axel-h, line 88:
```suggestion
/* Must be defined in virtio_vsock_emul.c */
```

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,243 @@
...
+ * to transmit the data to the guest VM.
+ *
+ * @param queue Receive queue
+ */
+static void virtio_vsock_notify_recv(virtqueue_device_t *queue)
+{
+    virtqueue_ring_object_t handle;
+
+    while (virtqueue_get_available_buf(queue, &handle)) {
+        char temp_buf[VIRTIO_VSOCK_CAMKES_MTU];
```
axel-h, line 105: This is tmp buffer is allocated on the stack. Might be nicer to make this part of `virtqueue_device_t`?

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,243 @@
...
+ * Sets up virtio vsock connections.
+ */
+static void make_virtio_vsock_virtqueues(void)
+{
+    int err;
+    for (int i = 0; i < NUM_PORTS; i++) {
+        virtqueue_driver_t *vq_send;
+        virtqueue_device_t *vq_recv;
+
+        vq_recv = malloc(sizeof(*vq_recv));
```
axel-h, line 168: Sometime we use `malloc()` and sometime `calloc()`. SHould we always use `calloc()`, as this gives us zero'd memory?

components/VM_Arm/src/modules/virtio_con.c
```diff
@@ -22,61 +42,160 @@
...
+ * Way of indexing: `head` and `tail` should not increase beyond (BUFSIZE - 1), as
+ * they repersent the real index in `buf`.
+ */
+typedef struct serial_shmem {
     uint32_t head;
     uint32_t tail;
     char buf[BUFSIZE];
-} extern volatile *batch_buf;
+} serial_shmem_t;
+compile_time_assert(serial_shmem_1k_size, sizeof(serial_shmem_t) == 0x1000);
```
axel-h, line 75:
```suggestion
compile_time_assert(serial_shmem_4k_size, sizeof(serial_shmem_t) == 0x1000);
```

components/VM_Arm/src/modules/virtio_con.c
```diff
@@ -22,61 +42,160 @@
...
+
+/**
+ * Initialize a virtio console backend. This includes:
+ * - adding an IO port handler
+ * - adding a PCI entry
+ * - initializing the emul layer
+ * - setting up connections (includes hvc0)
+ *
+ * @param vm The vm handler of the caller
+ * @param cookie Unused
```
axel-h, line 158: I'd still put in some dummy value here to force the caller to use this properly. Otherwise the callers might ignore this, and we have a potential problem in the future when this cookie is really needed. Practically, we could also put `connections` in here, as this seems our global context at the moment.

components/VM_Arm/src/modules/virtio_con.c
```diff
@@ -1,10 +1,30 @@
 /*
- * Copyright 2019, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2022, UNSW (ABN 57 195 873 179)
```
axel-h, line 2: I think the old copyright should be kept

components/VM_Arm/src/modules/virtio_con.c
```diff
@@ -22,61 +42,160 @@
...
+
+/**
+ * Writes data to the guest console attached to a port, and sets the head of the receive
+ * ring buffer to the right place.
+ *
+ * @param port Port number
+ */
+static void virtio_con_notify_recv(int port)
+{
+    volatile serial_shmem_t *buffer = connections[port].recv_buf;
```
axel-h, line 99: Could add a `assert((port >=0) && (port < ARRAY_SIZE(connections)));` here. Or simply pass a `serial_conn_t*` here instead of the port type, as thgis seems to be the context we are operating on. That avoids unnecessary dependencies on the global variables.

components/Init/src/virtio_con.c
```diff
@@ -0,0 +1,227 @@
...
+    buffer->head = buffer->tail;
+}
+
+/**
+ * Callback function for camkes virtqueue notification. Polls each receive ring buffer
+ * to see if there are data pending.
+ *
+ * @see main.c (struct device_notify) for parameters
+ */
+void handle_serial_console(vm_t *vmm)
```
axel-h, line 117: Maybe add a comment that for some legacy reason we get a `vmm` here, that we don't use. Where does it come from actually, should we have an assert that our connection VM has the same VMM than the VMM passed here?

components/VM_Arm/src/modules/virtio_con.c
```diff
@@ -22,61 +42,160 @@
...
+ */
+typedef struct serial_conn {
+    void (*notify)(void);
+    volatile serial_shmem_t *recv_buf;
+    volatile serial_shmem_t *send_buf;
+} serial_conn_t;
+static serial_conn_t connections[NUM_PORTS];
+
+/* defined in virtio_console_emul.c */
+void virtio_console_putchar(int port, virtio_emul_t *con, char *buffer, uint32_t head, uint32_t tail) WEAK;
```
axel-h, line 89: Can't we move this prototype definition to a header file? I also wonder, what does this do practically? It make some function implemented somewhere else weak, that seems a bit odd. The weak should be where the implementation is?

components/VM_Arm/src/modules/virtio_con.c
axel-h: Ah ok, we pass the port downstream again. But it's a bit odd, shouldn't there be a cookie then for the lower layer as reference, where we don't make a strong assumption what context this holds (might still be the plain port).

components/Init/src/virtio_con.c
axel-h: I think it must be volatile because this is shared memory. But I wonder if we need even stronger access constraints then with barriers, to really ensure the proper visibility. It might be worth moving the whole shares mem access to a dedicated file then that wraps all these details behind a nice API, that allows re-using it then also.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,243 @@
...
+            connections[i].send_queue->notify();
+            return;
+        }
+    }
+
+    ZF_LOGE("Invalid CID: %d", cid);
+}
+
+/**
+ * Sets up virtio vsock connections.
```
axel-h, line 159: Maybe add a note here that this calls `abort()` on any error, so there is no special error handling or recovery.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    vsock_cookie->vm = vm;
+    backend.vsock_data = (void *) vsock_cookie;
+
+    ioport_range_t virtio_port_range = {0, 0, VIRTIO_IOPORT_SIZE};
+    virtio_vsock = common_make_virtio_vsock(vm, pci, io_ports, virtio_port_range, IOPORT_FREE,
+                                            VIRTIO_SCK_IRQ, VIRTIO_SCK_IRQ, backend);
+
+    make_virtio_vsock_virtqueues();
```
axel-h, line 236: Maybe add a note here that this calls `abort()` on any error, so there is no special error handling or recovery implemented for now.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,243 @@
...
+    assert(virtio_vsock);
```
axel-h, line 237: please move the assert up. There seem not much point in calling `make_virtio_vsock_virtqueues()` when this fails.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,243 @@
...
+    virtio_vsock_cookie_t *vsock_cookie;
+
+    backend.guest_cid = guest_cid;
+    backend.injectIRQ = vsock_inject_irq;
+    backend.forward = tx_forward;
+
+    vsock_cookie = (virtio_vsock_cookie_t *)calloc(1, sizeof(virtio_vsock_cookie_t));
+    if (vsock_cookie == NULL) {
+        ZF_LOGE("Failed to allocated virtio vsock cookie");
+        return;
```
axel-h, line 226: This seems inconsistent with the other error behavior below. We should call `ZF_LOGF()`then also to abort. And the documentation should state that this is the designed error behavior, there is no fallback option in the current design. Otherwise this function should return an error code, so the caller can decide how to handle a failure.

components/Init/src/virtio_con.c
abrandnewusername: Can't think of the reason why the vm handler is there in the first place, since none of the implementations of this callback makes use of it. Maybe it was there for debugging purposes? I suggested the parameter should be removed in future work.

components/VM_Arm/src/modules/virtio_con.c
abrandnewusername: based on previous implementations of virtio backends, instead of calling the function `virtio_console_putchar` directly from this file, there should be an "emul driver" structure that contains a pointer to this function that got passed from the emul layer to this layer.
I'll make it so unless anyone has objections.

components/VM_Arm/src/modules/virtio_con.c
abrandnewusername: Is it necessary since the emul layer also has knowledge about the port structure?

components/Init/src/virtio_con.c
wom-bat: I'd prefer a different way of indexing.
  Always add at buf[++head % BUFSIZE]
  (typically add by memcpying into buf, then a memory barrier, then update head, so consumer can't see until all is added)
  Always take from [tail++ % BUFSIZE]
  empty = (head-tail)%BUFSIZE==0
  full = (head - tail) % BUFSIZE == BUFSIZE-1

wom-bat: Right now you have the invariants, 0 <= head < BUFSIZE
0 <= tail < BUFSIZE
empty = head - tail == 0
full = head > tail ? head - tail == BUFSIZE-1 : head + BUFSIZE-tail == BUFSIZE-1

components/Init/src/virtio_con.c
wom-bat: It's shared memory, yes, but it doesn't need 'volatile' if the memory barriers are correct, and 'volatile' may hurt performance.

components/Init/src/virtio_vsock.c
```diff
@@ -0,0 +1,247 @@
...
+        while (camkes_virtqueue_driver_gather_buffer(queue, &handle, &buf, &buf_size, &flag) >= 0) {
+            camkes_virtqueue_buffer_free(queue, buf);
+        }
+    }
+}
+
+/**
+ * Receives data from the source VM using camkes virtqueue, and invokes the emul layer
+ * to transmit the data to the guest VM.
+ *
```
wom-bat, line 91: Is this function invoked in the context of the source or the destination VM? Looks like it has to be the source VM.

components/Init/src/virtio_vsock.c
wom-bat: What's worse it enforces a double copy of the data, in line 101 and in the callback.

components/Init/src/virtio_vsock.c
abrandnewusername: Should be the dest VM. When the dest VMM receives a notification from the camkes virtqueue (comes from the source VMM), it invokes `handle_vsock_irq` which then invokes `virtio_vsock_notify_recv` to handle the data in the camkes virtqueue.

components/Init/src/virtio_con.c
abrandnewusername: What is the advantage of having ringbuffer indexes that increase continually? Seems like it can make the calculation easier, but you need to do a modular every time when accessing the buffer.

abrandnewusername: Also, the current structure of the ringbuffer is originally used by the [SerialServer](https://github.com/seL4/global-components/blob/master/components/SerialServer/src/serial.c#L42).
The arm version of the virtio console backend (see [here](https://github.com/seL4/camkes-vm/blob/ce6065f3caa230ef6c712e61190daebeb74fc005/components/VM_Arm/src/modules/virtio_con.c#L146)) needs to handle VM-VM transactions as well as VM-Serial transactions, and has only one single interface for both of them. 
I can only change the VM-VM ringbuffers to use the indexing you propose, but not the VM-Serial one.  Not sure if it would be a good idea to have two variants of ringbuffer in the same component as they need to be handled differently.

components/Init/src/virtio_vsock.c
abrandnewusername: This is due to the scatter-gather buffer that camkes virtqueue is using. To avoid copying, the emul layer needs to be aware of the structure of the data. Or, I can change the backend to use some other transport mechanisms.

components/Init/src/virtio_con.c
wom-bat: > What is the advantage of having ringbuffer indexes that increase continually? Seems like it can make the calculation easier, but you need to do a modular every time when accessing the buffer.

Main thing is it's much easier to check the invariants, and it's a lot easier to get right.  The corner cases otherwise keep having to check for overflow; generally speaking a branch structure is harder to check for correctness, and slower than a simple modulo.  That's even more true if we could make the BUFSIZ a power of two,when modulo becomes an 'and'

abrandnewusername:
Updates:
- Addressed some minor issues, added more comments
- Removed `volatile` from the shared buffer, added memory barriers

Major issues remaining:
- one extra copy between the vsock emul layer and the backend layer (see https://github.com/seL4/camkes-vm/pull/36#discussion_r1050310784). To fix this we either need to make the backend aware of camkes virtqueue or use another transporting mechanism, not sure if that's worth it.
- Ringbuffer indexing (see https://github.com/seL4/camkes-vm/pull/36#discussion_r1050306907). There was some previous discussion about modifying virtio console ring buffers (both VMM-VMM as well as VMM-SerialServer) to use another way of indexing. this can easily be done on VMM-VMM buffers but not VMM-SerialServer buffers (since camkes SerialServer is widely used and we need to change every instance). 

I reckon these can be future works.

components/Init/src/virtio_irq.h
```diff
@@ -0,0 +1,23 @@
...
+ * for each device.
+ *
+ * @todo: Having hardcoded numbers might become an issue as the list of virtual
+ * devices grows. Might be worth it to have an IRQ number allocator.
+ */
+#define VIRTIO_NET_IRQ              6
+#define VIRTIO_BLK_IRQ              7
+
+/* Number 8 is occupied by the RTC timer, do not use! */
+#define RTC_IRQ                     8
```
axel-h, line 20: It seems a bit odd to define `RTC_IRQ` here given the comment. Is there any code that really needs this define and expects it to come from this file?

components/Init/src/virtio_vsock.c
axel-h: Please address this, as memory allocation could fail. Calling `ZF_LOGF()` seems the best thing.now.

components/Init/src/virtio_irq.h
abrandnewusername: Thanks for pointing that out, did some clean-up for the rtc file(`mc146818rtc.c`).

abrandnewusername:
Updates:
- added minor fixes and more comments
- fixed more magic IRQ numbers

Why is [this file](https://github.com/seL4/camkes-vm/blob/master/components/Init/src/mc146818rtc.c) so messy but got merged anyways? Since I have made some changes to it in this PR, do I also need to clean it up?

wom-bat (approved):

PR #37, opened 2022-10-19
vm_arm: add option for custom device tree base

chrisguikema:
When generating a device tree, the default behavior is to use the kernel device tree, which is passed down through the bootinfo struct. This works for some platforms, however, there are platforms where the bootloader or toolchain updates modify the device tree. In these cases, the dts in the kernel is not a usable base for a generated device tree.

This commit adds an option "VmCustomDtbBase" which when enabled will ignore the kernel device tree, and instead pull the linux_image_config.dtb_base defined in the CAmkES configuration from the File Server. This based is used as the fdt_ori pointer during the fdt generation process.

This option provides an alternative to updating the kernel device tree, keeping that file from changing everytime a toolchain updates.

Signed-off-by: Chris Guikema <chris.guikema@dornerworks.com>

### DIFF
--- a/components/VM_Arm/configurations/vm.h
+++ b/components/VM_Arm/configurations/vm.h
@@ -78,6 +78,7 @@
         string initrd_name = "linux-initrd"; \
         string linux_bootcmdline = ""; \
         string linux_stdout = ""; \
+        string dtb_base_name = ""; \
     } linux_image_config; \
 
 
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -101,8 +101,10 @@ irq_server_t *_irq_server;
 vmm_pci_space_t *pci;
 vmm_io_port_list_t *io_ports;
 reboot_hooks_list_t reboot_hooks_list;
+
 #define PLAT_LINUX_DTB_SIZE 0x50000
 static char linux_gen_dtb_buf[PLAT_LINUX_DTB_SIZE];
+static char linux_gen_dtb_base_buf[PLAT_LINUX_DTB_SIZE];
 
 struct ps_io_ops _io_ops;
 
@@ -843,15 +845,28 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
     }
 
     if (!config_set(CONFIG_VM_DTB_FILE)) {
-        camkes_io_fdt(&(_io_ops.io_fdt));
-        void *fdt_ori = ps_io_fdt_get(&_io_ops.io_fdt);
-
+        void *fdt_ori;
         void *gen_fdt = linux_gen_dtb_buf;
         int size_gen = PLAT_LINUX_DTB_SIZE;
         int num_paths = 0;
         char **paths = NULL;
         if (camkes_dtb_get_node_paths) {
             paths = camkes_dtb_get_node_paths(&num_paths);
+        }
+
+        int dtb_fd = open(linux_image_config.dtb_base_name, 0);
+
+        /* If dtb_base_name is in the file server, grab it and use it as a base */
+        if (dtb_fd >= 0) {
+            size_t dtb_len = read(dtb_fd, linux_gen_dtb_base_buf, PLAT_LINUX_DTB_SIZE);
+            if (dtb_len <= 0) {
+                return -1;
+            }
+            close(dtb_fd);
+            fdt_ori = (void *)linux_gen_dtb_base_buf;
+        } else {
+            camkes_io_fdt(&(_io_ops.io_fdt));
+            fdt_ori = (void *)ps_io_fdt_get(&_io_ops.io_fdt);
         }
 
         err = generate_fdt(vm, fdt_ori, gen_fdt, size_gen, initrd_image.size, paths, num_paths);


components/VM_Arm/src/main.c
```diff
@@ -715,6 +717,34 @@ static int route_irqs(vm_vcpu_t *vcpu, irq_server_t *irq_server)
     return 0;
 }
 
+/* Different from load_image which loads the image into the guest's address space. This pulls the
+ * image from the FileServer and stores it into a local buffer
+ */
+static int get_file(const char *name, char *buf)
```
axel-h, line 723: This returns a length or 0 and not an error code, so the return type should be `size_t` then.
```suggestion
static size_t get_file(const char *name, char *buf)
```

components/VM_Arm/src/main.c
```diff
@@ -826,6 +854,17 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
             paths = camkes_dtb_get_node_paths(&num_paths);
         }
 
+#ifdef CONFIG_VM_CUSTOM_DTB_BASE
+        err = get_file(linux_image_config.dtb_base, linux_gen_dtb_base_buf);
+        if (!err) {
```
axel-h, line 859: It feels semantically wrong/misleading to use the expression `!err` here. Might be better to use 
```
int dtb_len = get_file(...);
if (0 == dtb_len) { ... error 
```

components/VM_Arm/src/main.c
```diff
@@ -715,6 +717,34 @@ static int route_irqs(vm_vcpu_t *vcpu, irq_server_t *irq_server)
...
+
+    fd = open(name, 0);
+    if (fd == -1) {
+        ZF_LOGE("Error: Unable to open image \'%s\'", name);
+        return 0;
+    }
+
+    size_t offset = 0;
+    while (1) {
+        /* Load the image BIT(20) bytes at a time */
```
axel-h, line 736: - Seems this function is a generic file loader after all. So there should also be a max buffer size parameter passed to have user from surprises when the file is really huger.
- Can you add a comment why BIT(20) is the limit here actually? It this a random choice or is this the shared memory limit that must be kept in mind here?

components/VM_Arm/configurations/vm.h
```diff
@@ -78,6 +78,7 @@
         string initrd_name = "linux-initrd"; \
         string linux_bootcmdline = ""; \
         string linux_stdout = ""; \
+        string dtb_base = ""; \
```
axel-h, line 81: maybe `dtb_base_name` or `dtb_name` is a better name choice here (and aligned with the other file name params) , so this can't be confused with a base address.

components/VM_Arm/CMakeLists.txt
```diff
@@ -127,6 +127,20 @@ config_option(
     OFF
 )
 
+config_option(
+    VmCustomDtbBase
```
axel-h, line 131: better use `VmCustomDtbBaseFile` and `VM_CUSTOM_DTB_BASE_FILE` to make it clear this is a file name.

components/VM_Arm/src/main.c
chrisguikema: BIT(20) is 1MB, and the `get_file` function was pulled from the seL4_projects_libs library, which pulled 1MB at a time. There really isn't a reason.

I'll add a max_size parameter. The library version just loads images to guest address space, so there isn't a worry about buffer overruns.

components/VM_Arm/configurations/vm.h
chrisguikema: Good idea. Updated to `dtb_base_name`.

components/VM_Arm/CMakeLists.txt
chrisguikema: Done.

components/VM_Arm/src/main.c
chrisguikema: Good call. I've updated the return value.

components/VM_Arm/src/main.c
```diff
@@ -715,6 +717,40 @@ static int route_irqs(vm_vcpu_t *vcpu, irq_server_t *irq_server)
     return 0;
 }
 
+/* Different from load_image which loads the image into the guest's address space. This pulls the
+ * image from the FileServer 1MB at a time and stores it into a local buffer
+ */
+static size_t get_file(const char *name, char *buf, int max_length)
```
axel-h, line 723: why not `size_t`for `max_length` also?

components/VM_Arm/src/main.c
```diff
@@ -715,6 +717,40 @@ static int route_irqs(vm_vcpu_t *vcpu, irq_server_t *irq_server)
...
+        ZF_LOGE("Error: Unable to open image \'%s\'", name);
+        return 0;
+    }
+
+    size_t offset = 0;
+    while (offset < max_length) {
+        /* Load the image 1MB at a time, until the file is read or
+         * the max_length parameter has been reached
+         */
+        len = read(fd, buf + offset, BIT(20));
```
axel-h, line 743: Need to take into account the remaining buffer size:
```suggestion
        len = read(fd, buf + offset, MIN(BIT(20), max_length - offset));
```

components/VM_Arm/CMakeLists.txt
```diff
@@ -127,6 +127,20 @@ config_option(
     OFF
 )
 
+config_option(
+    VmCustomDtbBaseFile
+    VM_CUSTOM_DTB_BASE_FILE
+    "Use a custom device tree base when generating a guest's device tree. This can be useful when \
+    the device tree needed by the guest is significantly different than the device tree provided by \
+    the kernel. For example, the TX2 bootloader modifies the device tree, so the tx2.dts in the \
```
axel-h, line 135: Maybe clarify this comment a bit more when it comes to the phrase "device tree provided by the kernel" and say "provided by the platform" . As far as I know, we compile the kernel against a given devices tree file. When booting in the platform, the device tree passed is the one from the platform bootloader, which can be different already to the one the kernel was compiled with. So the actual problem might start by a wrong device tree used during kernel compilation or missing sanity check on the ElfLoader or kernel, that things still match. What is the actual issue you are running into, is the memory configuration different because the boot loader carves out memory dynamically? In the worst case we might have much bigger problem on the TX2  platform then...?

chrisguikema: I'll make that change.

There's also the ZCU102, which in 2021.1 (?) updated the bus from `amba` to `axi`, so if you generate a device tree from the `zynqmp.dts` in the kernel while using a 2021.1 kernel, things won't work. But with this change you could keep the kernel unmodified and pass a new device tree to the VM to generate against.

As for the TX2, I was unaware (and maybe things have been updated since I worked with the platform) that the `tx2.dts` in the kernel is _not_ the device tree used as a base for the generation.

components/VM_Arm/src/main.c
chrisguikema: Nice catch. Will update.

components/VM_Arm/src/main.c
chrisguikema: Will update.

components/VM_Arm/src/main.c
```diff
@@ -715,6 +717,40 @@ static int route_irqs(vm_vcpu_t *vcpu, irq_server_t *irq_server)
...
 
+/* Different from load_image which loads the image into the guest's address space. This pulls the
+ * image from the FileServer 1MB at a time and stores it into a local buffer
+ */
+static size_t get_file(const char *name, char *buf, size_t max_length)
+{
+    int fd;
+    size_t len = 0;
+
+    if (max_length <= 0) {
```
axel-h, line 728: Let's drop  this check. It's always greater or equal to zero. Passing 0 runs into the the loop ad breaks there, so that handled well also

chrisguikema: Done.

components/VM_Arm/CMakeLists.txt
```diff
@@ -127,6 +127,18 @@ config_option(
     OFF
 )
 
+config_option(
```
axel-h, line 130: I wonder, does this have to be a config option or can we just make this feature available everywhere. We could check if the string `dtb_base_name` is set or not and then act accordingly.

components/VM_Arm/src/main.c
```diff
@@ -815,8 +847,6 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
     }
 
     if (!config_set(CONFIG_VM_DTB_FILE)) {
-        camkes_io_fdt(&(_io_ops.io_fdt));
```
kent-mcleod, line 818: I've seen similar changes to this in a patch by @WellMcGarnicle [1].  In that patch, instead of adding a new config option, the patch assumes that if a dtb_name attribute is set in the camkes config it refers to a file in the file server (as that's what the variable's purpose is). So if CONFIG_VM_DTB_FILE is not set, it tries to get the file from the file server but falls back to the camkes_io_fdt if the file doesn't exist. 

[1]
```
diff --git a/components/VM_Arm/src/main.c b/components/VM_Arm/src/main.c
index 2b14989..07ab026 100644
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -104,6 +104,7 @@ vmm_io_port_list_t *io_ports;
 reboot_hooks_list_t reboot_hooks_list;
 #define PLAT_LINUX_DTB_SIZE 0x50000
 static char linux_gen_dtb_buf[PLAT_LINUX_DTB_SIZE];
+static char linux_original_dtb_buf[PLAT_LINUX_DTB_SIZE];
 
 struct ps_io_ops _io_ops;
 
@@ -829,8 +830,23 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
     }
 
     if (!config_set(CONFIG_VM_DTB_FILE)) {
-        camkes_io_fdt(&(_io_ops.io_fdt));
-        void *fdt_ori = ps_io_fdt_get(&_io_ops.io_fdt);
+        void *fdt_ori = NULL;
+        // Check if we've been given a dtb to use as a base for generation.
+        int dtb_fd = open(dtb_name, 0);
+        if (dtb_fd == -1) {
+          printf("FDT Generation: Using camkes io dtb as base\n");
+          // No DTB found, using the camkes built in dtb.
+          camkes_io_fdt(&(_io_ops.io_fdt));
+          fdt_ori = ps_io_fdt_get(&_io_ops.io_fdt);
+        } else {
+          printf("FDT Generation: Using provided dtb ('%s') as base\n", dtb_name);
+          size_t len = read(dtb_fd, linux_original_dtb_buf, PLAT_LINUX_DTB_SIZE);
+          if (len <= 0) {
+            ZF_LOGE("Failed to read provided dtb");
+            return -1;
+          }
+          fdt_ori = (void *)linux_original_dtb_buf;
+        }
 
         void *gen_fdt = linux_gen_dtb_buf;
         int size_gen = PLAT_LINUX_DTB_SIZE;
```

chrisguikema: Seems really close to what I'm trying to achieve here. Do you want to pull that in, or should I update my PR to make the changes more similar to that?

components/VM_Arm/src/main.c
```diff
@@ -826,6 +856,19 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
             paths = camkes_dtb_get_node_paths(&num_paths);
         }
 
+        /* If dtb_base_name is in the file server, grab it and use it as a base */
+        if (open(linux_image_config.dtb_base_name, 0)) {
+            size_t dtb_len = get_file(linux_image_config.dtb_base_name, linux_gen_dtb_base_buf,
```
kent-mcleod, line 861: Can get_file just be replaced with: `size_t dtb_len = read(dtb_fd, linux_gen_dtb_base_buf, PLAT_LINUX_DTB_SIZE);`?
Using the fd returned by the open call above, and then followed with a close() after.

components/VM_Arm/src/main.c
kent-mcleod: Yea, for a second I thought they were the same implementation initially. It's fine to merge this one if you're also happy removing the config option.

components/VM_Arm/src/main.c
chrisguikema: Sure. That cleans up the commit nicely.

components/VM_Arm/src/main.c
```diff
@@ -854,6 +854,21 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
             paths = camkes_dtb_get_node_paths(&num_paths);
         }
 
+        int dtb_fd = open(linux_image_config.dtb_base_name, 0);
+
+        /* If dtb_base_name is in the file server, grab it and use it as a base */
+        if (dtb_fd) {
+            size_t dtb_len = read(dtb_fd, linux_gen_dtb_base_buf, PLAT_LINUX_DTB_SIZE);
+            if (0 == dtb_len) {
```
kent-mcleod, line 862: And `dtb_len <= 0`

chrisguikema: Done.

components/VM_Arm/src/main.c
```diff
@@ -854,6 +854,21 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
             paths = camkes_dtb_get_node_paths(&num_paths);
         }
 
+        int dtb_fd = open(linux_image_config.dtb_base_name, 0);
+
+        /* If dtb_base_name is in the file server, grab it and use it as a base */
+        if (dtb_fd) {
```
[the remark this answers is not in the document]
chrisguikema: Done.

kent-mcleod (approved):

PR #38, opened 2022-10-20
Device Tree Generation Fixes

chrisguikema:
Adds "vm-specific" plat_keep_devices and plat_keep_devices_and_subtree entries that can be defined in the devices.camkes file. Allows for more customization of device tree generation.

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -124,6 +124,8 @@ seL4_Error WEAK camkes_dtb_get_irq_cap(int irq, seL4_CNode cnode, seL4_Word inde
 simple_get_IRQ_handler_fn original_simple_get_irq_fn;
 int *WEAK camkes_dtb_get_irqs(int *num_irqs);
 char **WEAK camkes_dtb_get_node_paths(int *num_nodes);
+char **WEAK camkes_dtb_get_plat_keep_devices(int *num_nodes);
+char **WEAK camkes_dtb_get_plat_keep_devices_and_subtree(int *num_nodes);
 
 #ifdef CONFIG_ARM_SMMU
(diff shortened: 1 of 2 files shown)


chrisguikema:
I think this was something we've used internally, so I can't find any public examples approved for release :( 

But consider this: https://github.com/seL4/camkes-vm/blob/f70b3ffddca50f138cd478249e5d9dd403f57370/components/VM_Arm/plat_include/tx2/plat/vmlinux.h#L169

If there were 2 VMs, each would generate a device tree with all of these devices, which obviously won't work. This commit basically allows you to define that structure in the `devices.camkes`.

So some pseudocode:

```
vm0.plat_keep_devices = [ some devices ]

vm1.plat_keep_devices = [ other devices ]
```

components/VM_Arm/src/main.c
```diff
@@ -719,15 +721,42 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
...
 {
     int err = 0;
+
+    int num_keep_devices;
+    char **keep_devices;
+    if (camkes_dtb_get_plat_keep_devices) {
+        keep_devices = camkes_dtb_get_plat_keep_devices(&num_keep_devices);
+    }
+
+    int num_keep_devices_and_subtree;
```
kent-mcleod, line 731: num_keep_devices and num_keep_devices_and_subtree should be initialized to 0 in case camkes_dtb_get_plat_keep_devices is null and they don't get initialized.

chrisguikema: Good catch. Done.

kent-mcleod (approved):

PR #39, opened 2022-10-20
trivial: change vm naming scheme

chrisguikema:
VMs may not be Linux :)


axel-h (approved):
Yes, this change it something I have also been using to properly identify VMs.

kent-mcleod (approved):
Nice!

PR #40, opened 2022-10-20
vm_arm, trivial: align serial inputs to vm number

chrisguikema:
Sets getchar_clients to be their VM number.

Test with: seL4/global-components#40

### DIFF
--- a/components/VM_Arm/configurations/vm.h
+++ b/components/VM_Arm/configurations/vm.h
@@ -137,10 +137,13 @@
     time_server.simple = true;
 
 #define PER_VM_VIRTUAL_SERIAL_CONFIGURATION_DEF(num) \
+    vm##num.serial_getchar_attributes = num + 1; \
+    Vm##num.guest_putchar_attributes = num + 1; \
+    vm##num.putchar_attributes = num + 1; \
+    vm##num.batch_attributes = num + 1; \
     vm##num.serial_getchar_shmem_size = 0x1000; \
     vm##num.batch_shmem_size = 0x1000; \
 
 #define VM_VIRTUAL_SERIAL_CONFIGURATION_DEF(vm_ids...) \
     VM_VIRTUAL_SERIAL_GENERAL_CONFIGURATION_DEF() \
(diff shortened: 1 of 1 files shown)


axel-h (approved):

components/VM_Arm/configurations/vm.h
```diff
@@ -132,6 +132,7 @@
     time_server.simple = true;
 
 #define PER_VM_VIRTUAL_SERIAL_CONFIGURATION_DEF(num) \
+    vm##num.serial_getchar_attributes = num; \
```
kent-mcleod, line 135: This same approach could be used for the output channels to ensure that the guests are always assigned the same colors?

chrisguikema: It could, yeah. But I haven't seen that behavior like I was seeing with the `getchar`.

chrisguikema:
@kent-mcleod is this good to merge?

kent-mcleod:
> @kent-mcleod is this good to merge?

This is coupled to https://github.com/seL4/global-components/pull/40, so is currently blocked by it.

axel-h:
> This is coupled to https://github.com/seL4/global-components/pull/40, so is currently blocked by it.

This is merged now, so anything still blocking the merge now?

lsf37 (approved):

kent-mcleod:
> This is merged now, so anything still blocking the merge now?

I recall I tried to explain [somewhere](https://github.com/seL4/global-components/pull/40#discussion_r1003651136) that using the VM ID as the serial sever client ID on the interfaces wasn't the right approach hence why I hadn't approved it.

lsf37:
Ah, sorry, was a bit too quick on this one. Should we revert?

kent-mcleod:
> Ah, sorry, was a bit too quick on this one. Should we revert?

Yes I think so.  https://github.com/seL4/global-components/pull/45 is what this PR was associated with.

lsf37:
Ok, will push a revert.

lsf37:
Reverted in d45e30b194b.

I don't think there is an option to re-open the PR, so we'll have to raise a new one when it is figured out what should be done.

PR #41, opened 2022-10-24
CAmkES ARM VM trivial fixes

JoonasOnatsu:
This PR contains trivial code style/formatting fixes and removal of hardcoded constants.


components/VM_Arm/src/main.c
```diff
@@ -94,7 +95,8 @@ simple_t _simple;
 vspace_t _vspace;
 sel4utils_alloc_data_t _alloc_data;
 allocman_t *allocman;
-static char allocator_mempool[83886080];
+#define ALLOCATOR_MEMPOOL_SZ ((size_t)80 * 1024 * 1024) /* 80 MB*/
+static char allocator_mempool[ALLOCATOR_MEMPOOL_SZ];
```
axel-h, line 99:
```suggestion
static char allocator_mempool[80*1024*1024];
```
and then using `sizeof ()` avoids the need to have an explicit define here. Seems we do not really need it at them moment and the compiler can handle this well. We could add the define once we make this configurable via CMake.

components/VM_Arm/src/main.c
```diff
@@ -188,11 +188,11 @@ static int _dma_morecore(size_t min_size, int cached, struct dma_mem_descriptor
     dma_desc->vaddr = _vaddr;
     dma_desc->paddr = getaddr_ret.paddr;
     dma_desc->cached = 0;
-    dma_desc->size_bits = 12;
+    dma_desc->size_bits = seL4_PageBits;
     dma_desc->alloc_cookie = (void *)frame;
     dma_desc->cookie = NULL;
     /* Advance the virtual address marker */
-    _vaddr += BIT(12);
+    _vaddr += seL4_PageBits;
```
kent-mcleod, line 195: `BIT(seL4_PageBits)`?

components/VM_Arm/src/main.c
chrisguikema: This mempool is tricky. If its not big enough, you can't allocate large amounts of RAM to a guest. That leads to a problem where you have two VMs, one with a large ram need, and one with a small ram need, and you're wasting resources by needing the mempool to be large enough to the high-ram-usage VM.

kent-mcleod: This should really be configurable from the CAmkES component configuration.

chrisguikema: If someone makes a PR for that, assign me as a reviewer. I did some testing and got a formula for the allocator_mempool -> ram_size relationship.

chrisguikema: Took care of it myself: https://github.com/seL4/camkes-vm/pull/45/

Smattr: This PR does not affect this behaviour, but shouldn't this pool attempt to align itself more strongly somehow? [allocman seems to deal with unaligned pools](https://github.com/seL4/seL4_libs/blob/0aa8d43fe53564e40248215d2b3b5cedddcc1ee3/libsel4allocman/src/bootstrap.c#L86)¹, but this component should probably take steps to make allocman’s fixup a no-op. This is likely not visible under GCC, which will probably 4K-align this array anyway. Maybe you want `__attribute__((aligned(4096)))`?

¹ Though `sizeof(size_t)` doesn’t seem like the right thing to use here.

kent-mcleod: Does the mspace_pool require a 4k alignment?  I didn't think it did, but if it does then we should reject unaligned init args.

axel-h: > Took care of it myself: #45

That looks nice, I think we should pick this up.
@JoonasOnatsu: could you drop the array changes from this PR, so we have no merge conflicts?

components/VM_Arm/src/main.c
JoonasOnatsu: Thanks for spotting this, seems like my tired eyes missed something. Will fix.

components/VM_Arm/src/main.c
JoonasOnatsu: > > Took care of it myself: #45
> 
> That looks nice, I think we should pick this up. @JoonasOnatsu: could you drop the array changes from this PR, so we have no merge conflicts?

Will do, that other solution seems much nicer anyways.

components/VM_Arm/src/main.c
JoonasOnatsu: Fixed now.

components/VM_Arm/src/main.c
JoonasOnatsu: I dropped the change on the latest push.

kent-mcleod (approved):

PR #42, opened 2022-10-24
Zynqmp vmm

chrisguikema:
Add support for compiling a zynqmp VMM. Example projects can be added to camkes-vm-examples once the SMC RFC has been worked through. This just adds the structure for the zynqmp.

### DIFF
--- a/arm_vm_helpers.cmake
+++ b/arm_vm_helpers.cmake
@@ -89,6 +89,8 @@ function(DeclareCAmkESARMVM init_component)
         set(vm_plat_include "${ARM_VM_PROJECT_DIR}/components/VM_Arm/plat_include/exynos5410")
     elseif(KernelPlatformExynos5422)
         set(vm_plat_include "${ARM_VM_PROJECT_DIR}/components/VM_Arm/plat_include/exynos5422")
+    elseif(KernelPlatformZynqmpUltra96v2)
+        set(vm_plat_include "${ARM_VM_PROJECT_DIR}/components/VM_Arm/plat_include/ultra96v2")
     else()
         set(
             vm_plat_include
--- a/arm_vm_helpers.cmake
+++ b/arm_vm_helpers.cmake
@@ -51,6 +51,10 @@ function(DeclareCAmkESARMVM init_component)
     list(APPEND vm_src ${ARM_VM_PROJECT_DIR}/components/VM_Arm/src/modules/map_frame_hack.c)
     list(APPEND vm_src ${ARM_VM_PROJECT_DIR}/components/VM_Arm/src/modules/init_ram.c)
 
+    if(VmVirtUart)
+        list(APPEND vm_src ${ARM_VM_PROJECT_DIR}/components/VM_Arm/src/modules/vuart_init.c)
+    endif()
+
     if(Tk1DeviceFwd)
         list(
             APPEND vm_src ${ARM_VM_PROJECT_DIR}/components/VM_Arm/src/modules/plat/tk1/device_fwd.c
--- a/components/VM_Arm/CMakeLists.txt
+++ b/components/VM_Arm/CMakeLists.txt
@@ -32,7 +32,7 @@ config_option(
     VM_PCI_SUPPORT
     "Enable virtual pci device support"
     DEPENDS
-    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR KernelPlatformTx2 OR KernelPlatformQEMUArmVirt OR KernelPlatformOdroidc2"
+    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR KernelPlatformTx2 OR KernelPlatformQEMUArmVirt OR KernelPlatformOdroidc2 OR KernelPlatformZynqmp"
     DEFAULT
     OFF
 )
@@ -52,7 +52,7 @@ config_option(
     VM_VIRTIO_NET_VIRTQUEUE
     "Enable virtio net virtqueue forwarding module"
     DEPENDS
-    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR KernelPlatformTx2 OR KernelPlatformQEMUArmVirt;VmPCISupport"
+    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR KernelPlatformTx2 OR KernelPlatformQEMUArmVirt OR KernelPlatformZynqmp;VmPCISupport"
     DEFAULT
     OFF
 )
--- a/components/VM_Arm/CMakeLists.txt
+++ b/components/VM_Arm/CMakeLists.txt
@@ -127,6 +127,16 @@ config_option(
     OFF
 )
 
+config_option(
+    VmVirtUart
+    VM_VIRT_UART
+    "Option for virtualizing console"
+    DEFAULT
+    OFF
+    DEPENDS
+    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR KernelPlatformZynqmp"
+)
+
 add_config_library(arm_vm "${configure_string}")
 
 DeclareCAmkESARMVM(VM)
--- a/components/VM_Arm/plat_include/ultra96v2/plat/vmlinux.h
+++ b/components/VM_Arm/plat_include/ultra96v2/plat/vmlinux.h
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2019, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#define GIC_NODE_PATH "/amba_apu@0/interrupt-controller@f9010000"
+
+static const int linux_pt_irqs[] = {};
+
+static const int free_plat_interrupts[] =  { -1 };
+
+static const char *plat_keep_devices[] = {};
+static const char *plat_keep_device_and_disable[] = {};
+static const char *plat_keep_device_and_subtree[] = {};
+static const char *plat_keep_device_and_subtree_and_disable[] = {};
+static const char *plat_linux_bootcmdline = "";
+static const char *plat_linux_stdout = "";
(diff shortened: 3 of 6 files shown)


components/VM_Arm/CMakeLists.txt
```diff
@@ -127,6 +127,16 @@ config_option(
...
 )
 
+config_option(
+    VmVirtUart
+    VM_VIRT_UART
+    "Option for virtualizing console"
+    DEFAULT
+    OFF
+    DEPENDS
+    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR  KernelPlatformZynqmp"
```
axel-h, line 137:
```suggestion
    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR KernelPlatformZynqmp"
```

chrisguikema: Ah dang, I hate when there's extra whitespace changes like that. I'll fix it.

components/VM_Arm/CMakeLists.txt
```diff
@@ -127,6 +127,16 @@ config_option(
     OFF
 )
 
+config_option(
+    VmVirtUart
```
axel-h, line 131: I wonder if we should call this `VmVirtConsole` and `VM_VIRT_CONSOLE`, as this seem the primary use case for this.

chrisguikema: While console is the obvious use case, its just a paravirtualized uart...so it doesn't have to be a console. Also, there's already a `VM_VIRTIO_CONSOLE`, so I think that would be confusing.

components/VM_Arm/src/modules/vuart_init.c
```diff
@@ -0,0 +1,41 @@
...
+
+#include <vmlinux.h>
+#include <camkes.h>
+
+#include <sel4vmmplatsupport/plat/devices.h>
+
+extern void *serial_getchar_buf;
+
+//this matches the size of the buffer in serial server
+#define BUFSIZE 4088
```
axel-h, line 15: Can we use some shares header structure here, so this is kept in sync? Is there a chance to rename this `serial_shared_page` instead of `serial_getchar_buf`. Otherwise a comment here would be nice that explains where the 4088 byte come from.

chrisguikema: This is identical to the `virtio_console` module. Can we track an issue to create a shared header for a serial server interface?

components/VM_Arm/plat_include/ultra96v2/plat/vmlinux.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2019, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#define GIC_IRQ_PHANDLE 0x04
```
axel-h, line 9: Is there a way to get this from the device tree parser? If this can't be implemented now, a comment would be nice here and in the other files that copies the section from the device tree, so this is a bit more copy/paste save. We have seen on different qemu-arm-virt versions that the value can change - which causes some strange errors then.

libs/libvirtio/plat_include/zynqmp/virtio/virtio_plat.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2019, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#define IRQ_SPI_OFFSET 32
```
axel-h, line 9: Come comments here would be nice to explain where this values come from.

components/VM_Arm/plat_include/ultra96v2/plat/vmlinux.h
chrisguikema: That would be really nice. We actually had an issue with vpci wasn't working because this value didn't match what was in the rest of the device tree.

This parameter is defined for each platform, so I would track this as a separate issue. But it should be possible to read the gic node to get this value, that way we can avoid needing to define this.

libs/libvirtio/plat_include/zynqmp/virtio/virtio_plat.h
chrisguikema: This is standard for ARMv8 and is in the other platform specific virtio_plat.h headers.

@yyshen might have more clarification, but my understanding is that each processor has 32 Private Peripheral Interrupts (PPI). SPI (Shared Peripheral Interrupts) start at 32, hence the offset.

axel-h: I couldn't mark all three lines. The `IRQ_SPI_OFFSET` is somehow clear (could still move to a shared header eventually), but the `61` that is used twice could get a comment. Is this a random pick or is there a spec?

chrisguikema: Honestly I have no idea where that came from. If you do the `61 + 32` that's actually a GEM2 interrupt, which is an unused hardware device. The value should probably be updated to not interfere with a physical device. I'll have to test.

components/VM_Arm/src/modules/vuart_init.c
chrisguikema: Thanks, Axel!

libs/libvirtio/plat_include/zynqmp/virtio/virtio_plat.h
chrisguikema: I tested `158`, which doesn't correspond to a physical interrupt. I'll add a comment explaining the value

yyshen: yes, the first 32 IRQs are reserved for PPI and SGI. The IRQs for peripherals start from 32.

yyshen (approved):

libs/libvirtio/plat_include/zynqmp/virtio/virtio_plat.h
axel-h: Cab you add a comment in the code why the same interrupt us used for both channels? Is this driven by implementation needs?

chrisguikema: That is the case for each platform with VPCI support. Looking at the vpci generation code, that doesn't seem to be a requirement. So I thought it would be best to just match the behavior of previously supported platforms. We can track an issue to make each virtio device have its own interrupt.

chrisguikema:
@yyshen is this good to merge?

axel-h (approved):
Looks ok from my side

axel-h:
Together with https://github.com/seL4/camkes-vm-images/pull/6, this should allow a CI run of the minimal VM example on ZynqMP hardware (zcu102) then, right?

chrisguikema:
Technically, yes. However, the zynqmp uses a host of SMC calls during the boot process.

```
            0xc2000001, // PM_GET_API_VERSION
            0xc200000D, // PM_REQ_NODE
            0xc200000E, // PM_RELEASE_NODE
            0xc200000F, // PM_SET_REQUIREMENT
            0xc2000015, // PM_INIT_FINALIZE
            0xc2000017, // PM_FPGA_GET_STATUS
            0xc2000011, // PM_RESET_ASSERT
            0xc2000012, // PM_RESET_GET_STATUS
            0xc2000018, // PM_GET_CHIPID
            0xc200001C, // PM_PINCTRL_REQUEST
            0xc200001D, // PM_PINCTRL_RELEASE
            0xc200001F, // PM_PINCTRL_SET_FUNCTION
            0xc2000020, // PM_PINCTRL_CFG_PARAM_GET
            0xc2000021, // PM_PINCTRL_CFG_PARAM_SET
            0xc2000022, // PM_IOCTL
            0xc2000023, // PM_QUERY_DATA
            0xc2000024, // PM_CLOCK_ENABLE
            0xc2000025, // PM_CLOCK_DISABLE
            0xc2000026, // PM_CLOCK_GETSTATE
            0xc2000027, // PM_CLOCK_SETDIVIDER
            0xc2000028, // PM_CLOCK_GETDIVIDER
            0xc200002C, // PM_CLOCK_GETPARENT
            0xc2000A03, // PM_GET_TRUSTZONE_VERSION
            0x82001000, // IPI_MAILBOX_OPEN
            0x82001001, // IPI_MAILBOX_RELEASE
            0x82001005  // IPI_MAILBOX_ENABLE_IRQ
```

DW created an RFC to get community discussion on how we can best handle this. So I could create a vm_minimal example, and it would compile, and it would run up to the point of needing to make its first SMC call before dying.

My purpose with creating these PRs is to get the framework in place, such that when the SMC RFC is closed, an example can be created immediately, and we don't need to wait for these extra pieces to be put in place.

chrisguikema:
Alternatively, we could release a FreeRTOS or RTEMS image, but that mostly relies upon my "Mulitple OS" PRs going through, and I think those will have a longer review process.

chrisguikema:
@yyshen can this be merged?

arm_vm_helpers.cmake
```diff
@@ -89,6 +89,8 @@ function(DeclareCAmkESARMVM init_component)
         set(vm_plat_include "${ARM_VM_PROJECT_DIR}/components/VM_Arm/plat_include/exynos5410")
     elseif(KernelPlatformExynos5422)
         set(vm_plat_include "${ARM_VM_PROJECT_DIR}/components/VM_Arm/plat_include/exynos5422")
+    elseif(KernelPlatformZynqmpUltra96v2)
+        set(vm_plat_include "${ARM_VM_PROJECT_DIR}/components/VM_Arm/plat_include/ultra96v2")
```
kent-mcleod, line 97: Is the zynqmp platform path deliberately missing from here?

chrisguikema: `zynqmp` is covered by the default `else` conditional below. Ulttra96v2 also has `KernelPlatform` be `zynqmp`, so it needs to be handled differently.

kent-mcleod (approved):

PR #43, opened 2022-10-24
trivial: add debug prints for generate_fdt

chrisguikema:


components/VM_Arm/src/main.c
```diff
@@ -768,36 +769,42 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
     err = fdtgen_generate(context, fdt_ori);
     fdtgen_free_context(context);
     if (err) {
+        ZF_LOGE("Couldn't generate fdt_ori\n");
```
axel-h, line 772: I wonder, are the any error codes in use besides `-1`? Might be worth printing the error value then also to get a bit more insight what failed. Otherwise the error codes are lost.

chrisguikema: Sure, I'll add the error to the prints.

axel-h (approved):
Thanks, looks good to me.

chrisguikema:
@kent-mcleod can this be merged?

PR #45, opened 2022-10-24
vm_arm: add allocator mempool template

chrisguikema:
The size of the allocator_mempool determines, along with the number of untyped objects, how much RAM a VM can map. There is also a circumstance on a 2VM system where VM0 needs to map a lot of RAM, and VM1 needs to map a little. The current setup would require the user to increase the allocator_mempool buffer for each VM, wasting system resources.

This commit adds a new template that defines the allocator_mempool buffer. Simply put "vm0.allocator_mempool_size = X" in the configuration to set the mempool size. This allows each VM to configure the mempool for its own needs, which more efficiently utilizes system resources.

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -94,7 +94,6 @@ simple_t _simple;
 vspace_t _vspace;
 sel4utils_alloc_data_t _alloc_data;
 allocman_t *allocman;
-static char allocator_mempool[83886080];
 seL4_CPtr _fault_endpoint;
 irq_server_t *_irq_server;
 
@@ -435,7 +434,7 @@ static int vmm_init(void)
                    simple_get_cnode_size_bits(simple),
                    simple_last_valid_cap(simple) + 1 + num_extra_frame_caps,
                    BIT(simple_get_cnode_size_bits(simple)),
-                   sizeof(allocator_mempool), allocator_mempool
+                   get_allocator_mempool_size(), get_allocator_mempool()
                );
     assert(allocman);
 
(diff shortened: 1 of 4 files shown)


yyshen (approved):

templates/seL4AllocatorMempool.template.c
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2022, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <camkes.h>
+
+/*- set mempool_size = macros.ROUND_UP(configuration[me.name].get('allocator_mempool_size', 83886080), macros.PAGE_SIZE) -*/
```
axel-h, line 9: can you make this `80*1024*1024` to make it a bit nicer?

components/VM_Arm/src/main.c
```diff
@@ -134,6 +133,9 @@ seL4_CPtr camkes_get_smmu_cb_cap();
 seL4_CPtr camkes_get_smmu_sid_cap();
 #endif
 
+char *get_allocator_mempool(void);
```
axel-h, line 136: Could this prototypes move to a template header, so this is re-usable for other implementations also?

templates/seL4AllocatorMempool.template.c
chrisguikema: Sure

components/VM_Arm/src/main.c
chrisguikema: Sure. Just tested and I can add `TEMPLATE_HEADERS` to add a template header to `camkes.h`.

axel-h (approved):
Looks good to me.

I wonder where we could go from here. At the moment this template is basically providing a generic buffer. We could move the allocaman init code over also, so this could provide a `allocman_t *get_simple allocman()` to provide an allocator instead of the raw buffer...?

kent-mcleod (approved):
Awesome

PR #47, opened 2022-10-25
vm_arm: dereference linux

chrisguikema:
The VMM can run other Operating Systems, so referencing Linux isn't necessarily correct.

Test with: https://github.com/seL4/camkes-vm-examples/pull/30

### DIFF
--- a/components/VM_Arm/configurations/vm.h
+++ b/components/VM_Arm/configurations/vm.h
@@ -64,22 +64,22 @@
     attribute int num_extra_frame_caps; \
     attribute int extra_frame_map_address; \
     attribute { \
-        string linux_ram_base; \
-        string linux_ram_paddr_base; \
-        string linux_ram_size; \
-        string linux_ram_offset; \
+        string ram_base; \
+        string ram_paddr_base; \
+        string ram_size; \
+        string ram_offset; \
         string dtb_addr; \
         string initrd_max_size; \
         string initrd_addr; \
-    } linux_address_config; \
+    } vm_address_config; \
     attribute { \
-        string linux_name = "linux"; \
+        string kernel_name = "linux"; \
         string dtb_name = "linux-dtb"; \
         string initrd_name = "linux-initrd"; \
-        string linux_bootcmdline = ""; \
-        string linux_stdout = ""; \
+        string kernel_bootcmdline = ""; \
+        string kernel_stdout = ""; \
         string dtb_base_name = ""; \
-    } linux_image_config; \
+    } vm_image_config; \
 
 
 #define VM_COMPONENT_DEF(num) \
@@ -139,4 +139,3 @@
 #define VM_VIRTUAL_SERIAL_CONFIGURATION_DEF(vm_ids...) \
     VM_VIRTUAL_SERIAL_GENERAL_CONFIGURATION_DEF() \
     __CALL(PER_VM_VIRTUAL_SERIAL_CONFIGURATION_DEF, vm_ids) \
-
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -110,10 +110,10 @@ struct ps_io_ops _io_ops;
 
 static jmp_buf restart_jmp_buf;
 
-unsigned long linux_ram_base;
-unsigned long linux_ram_paddr_base;
-unsigned long linux_ram_size;
-unsigned long linux_ram_offset;
+unsigned long ram_base;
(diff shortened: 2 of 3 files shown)


components/VM_Arm/configurations/vm.h
```diff
@@ -64,22 +64,21 @@
...
         string initrd_addr; \
-    } linux_address_config; \
+    } vm_address_config; \
     attribute { \
-        string linux_name = "linux"; \
+        string kernel_name = "linux"; \
         string dtb_name = "linux-dtb"; \
         string initrd_name = "linux-initrd"; \
-        string linux_bootcmdline = ""; \
-        string linux_stdout = ""; \
```
kent-mcleod, line 80: This will break a lot of existing camkes systems which use these settings. I'm not sure if this change is worth it right now?

chrisguikema: The specific linux_bootcmdline -> kernel_bootcmdline change? Or the fact that everything was renamed?

It was a simple find/replace in camkes-vm-examples FWIW

kent-mcleod: > It was a simple find/replace in camkes-vm-examples FWIW

I mean the overall rename. Other systems that are implemented outside of camkes-vm-examples would also need to be updated including ones that aren't maintained by us.

chrisguikema: That's very true. I fixed the `camkes-vm-examples`, but if you link to others, I can make appropriate PRs.

But can't you make the same argument for any API change? Other repos that reference the changed repos need to make appropriate changes when updating. Maybe this would be good for an RFC? I think it ties into the concept of a "General VMM" that was discussed at the summit.

kent-mcleod: Yea, it's a breaking API change.  But the problem that this is addressing - that the names are confusing in a non-linux context - is still arguably a smaller problem than the new problem the change would create - that previously working applications won't compile and will need updating.

Instead, we could track the current problem in an issue, and when we're making a larger set of changes for supporting non-Linux VMs, refactor it then.

So my point is that the bad names isn't quite a bad enough problem to justify a breaking API change. So we should wait until the problem is large enough or there is another large breaking change already and make the change then.

chrisguikema: That's a good point.

I have another commit that handles some things on a "per-VM" level, instead of on a "per-project" level. For example, right now if you give a VM a DTB, it either needs to be provided to each VM, or generated for each VM. But you could have an RTOS that doesn't need a device tree, so the solution is to add a boolean for each of these instances in the VM configuration.

```
+        int provide_dtb = true; \
+        int generate_dtb = false; \
+        int provide_initrd = true; \
+        int clean_cache = false; \
+        int map_unity = false; \
     } vm_image_config; \
```

Basically, what we've noticed is that when you're supporting different Operating Systems, these things aren't always required. RTEMS requires a cache cleaning when loading the images into RAM because the guest OS isn't smart enough perform the cache clean.

If I created a PR for supporting multiple Operating Systems, is that a big enough change to cause an API break?

chrisguikema: Another example would be guest entry addresses. Linux requires the offset, but other Operating Systems might not.

chrisguikema: Or should that be an RFC, given how many projects that change would effect?

kent-mcleod: > I have another commit that handles some things on a "per-VM" level, instead of on a "per-project" level.

Yea, this limitation has been a bit frustrating. The `config_option()`'s in camkes-vm should be able to be set per-vm instead of per-project. Reworking this with additional fixes like how the ram device is specified and installed and how the boot images are installed into the ram device would all be sufficient improvements to warrant the API change IMO. (Although if there is an easy way to maintain backwards compatibility then we could still consider it).

chrisguikema: Alright, I'll make a bigger PR that the community can review to see if it warrants an API change. Thanks for the feedback. Personally, I think all the changes make for a much more usable VM!

Also, when I make the changes in `camkes-vm-examples`, I can only see if things compile, but each configuration should be tested either in simulation or on hardware to ensure it still boots.

chrisguikema: @kent-mcleod I made a new PR: https://github.com/seL4/camkes-vm/pull/48

PR #48, opened 2022-10-25
Multiple OS Support

chrisguikema:
Test with: https://github.com/seL4/seL4_projects_libs/pull/81
Test with: https://github.com/seL4/camkes-vm-examples/pull/31

### DIFF
--- a/components/VM_Arm/configurations/vm.h
+++ b/components/VM_Arm/configurations/vm.h
@@ -75,6 +75,7 @@
         string dtb_addr; \
         string initrd_max_size; \
         string initrd_addr; \
+        string kernel_entry_addr = "-1"; \
     } vm_address_config; \
     attribute { \
         string kernel_name = "linux"; \
@@ -83,6 +84,11 @@
         string kernel_bootcmdline = ""; \
         string kernel_stdout = ""; \
         string dtb_base_name = ""; \
+        int provide_dtb = true; \
+        int generate_dtb = false; \
+        int provide_initrd = true; \
+        int clean_cache = false; \
+        int map_one_to_one = false; \
     } vm_image_config; \
     attribute { \
         string linux_ram_base; \
--- a/components/VM_Arm/configurations/vm.h
+++ b/components/VM_Arm/configurations/vm.h
@@ -71,9 +71,7 @@
         string ram_base; \
         string ram_paddr_base; \
         string ram_size; \
-        string ram_offset; \
         string dtb_addr; \
-        string initrd_max_size; \
         string initrd_addr; \
         string kernel_entry_addr = "-1"; \
     } vm_address_config; \
--- a/components/VM_Arm/configurations/vm.h
+++ b/components/VM_Arm/configurations/vm.h
@@ -68,6 +68,23 @@
     attribute int num_extra_frame_caps; \
     attribute int extra_frame_map_address; \
     attribute { \
+        string ram_base; \
+        string ram_paddr_base; \
+        string ram_size; \
+        string ram_offset; \
+        string dtb_addr; \
+        string initrd_max_size; \
+        string initrd_addr; \
+    } vm_address_config; \
+    attribute { \
+        string kernel_name = "linux"; \
+        string dtb_name = "linux-dtb"; \
+        string initrd_name = "linux-initrd"; \
+        string kernel_bootcmdline = ""; \
+        string kernel_stdout = ""; \
+        string dtb_base_name = ""; \
+    } vm_image_config; \
+    attribute { \
         string linux_ram_base; \
         string linux_ram_paddr_base; \
         string linux_ram_size; \
--- a/components/VM_Arm/docs/vm_memory_configuration.md
+++ b/components/VM_Arm/docs/vm_memory_configuration.md
@@ -1,0 +1,85 @@
+<!--
+     Copyright 2023, DornerWorks
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# VM Memory Configuration
+
+This document overviews how the seL4 VMM provides guests with the a block of memory.
+
+## Basic Configuration
+
+In the platform specific `devices.camkes` file, there are macros that define the VM's RAM region:
+
+```
+#define VM_RAM_BASE 0x40000000
+#define VM_RAM_SIZE 0x20000000
+
+vm0.vm_address_config = {
+    "ram_base" : VAR_STRINGIZE(VM_RAM_BASE),
+    "ram_paddr_base" : VAR_STRINGIZE(VM_RAM_BASE),
+    "ram_size" : VAR_STRINGIZE(VM_RAM_SIZE),
+};
+
+vm0.simple_untyped24_pool = 33;
+
+```
+
+This configuration lets VM0 see 512MB at 0x40000000. When running as a hypervisor, seL4 uses stage 2
+mappings to map an Intermediate Physical Address (IPA) to a Physical Address (PA). This means that
+0x40000000 is the IPA that the guest sees, and the associated PA could be any available page in the
+system. The pages backing the guest mappings are provided by the `simple_untyped24_pool`. In order
+to map 512MB, 32 untyped objects of size 24 are required. The additional object is provided for
+overhead operations.
+
+### What does this mean?
+
+This default method is the most secure way of configuring the system. You can give two VMs the exact
+same `RAM_BASE` and `RAM_SIZE`, and the MMU will protect against the VMs accessing the others
+memory.
+
+The only case where this fails is when you want to give a VM a DMA-backed hardware device. In this
+specific case, you must provide the guest with the appropriate SMMU capabilities such that the SMMU
+will back the VM's DMA transactions.
+
+However, there are platforms that don't have a (supported) SMMU. What then?
+
+## 1:1 Mapping Configuration
+
+DMA transactions will still work in the guest if the IPA->PA mappings are 1:1. In the previous
+example, a VM at 512MB at 0x40000000. In order for DMA to work, the VMM must use the page
+capabilities for the specified address range to back the guests mappings. For example:
+
+```
+#define VM_RAM_BASE 0x40000000
+#define VM_RAM_SIZE 0x20000000
+
+vm0.vm_address_config = {
+    "ram_base" : VAR_STRINGIZE(VM_RAM_BASE),
+    "ram_paddr_base" : VAR_STRINGIZE(VM_RAM_BASE),
+    "ram_size" : VAR_STRINGIZE(VM_RAM_SIZE),
+};
+
+
+vm0.vm_image_config = {
+    "map_one_to_one" : true,
+};
+
+vm0.untyped_mmios = [
+    "0x40000000:29", // Linux kernel memory regions
+];
+
+vm0.simple_untyped24_pool = 1;
+
+```
+
+In this example, the capability to the guest memory region is provided via the `untyped_mmios`
+configuration option. The VMM knows to map 1:1 by the 'vm_image_config.map_one_to_one flag'. Also,
+the `simple_untyped24_pool` was reduced to 1 because the VMM no longer needs the generic untyped
+objects to back the guest's mappings.
+
+### How to get the Capability List
+
+Use the `CapDLLoaderPrintUntypeds` CMake option to have the capDL loader print out the untyped
+objects list. Then place the objects associated with the guest memory region into the
+`untyped_mmios` pool.
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -110,14 +110,6 @@ struct ps_io_ops _io_ops;
 
 static jmp_buf restart_jmp_buf;
 
-unsigned long linux_ram_base;
-unsigned long linux_ram_paddr_base;
-unsigned long linux_ram_size;
-unsigned long linux_ram_offset;
-unsigned long dtb_addr;
-unsigned long initrd_max_size;
-unsigned long initrd_addr;
-
 void camkes_make_simple(simple_t *simple);
 
 int WEAK camkes_dtb_untyped_count();
@@ -456,7 +448,7 @@ static int vmm_init(void)
         vka_cspace_make_path(vka, cap, &path);
         int utType = device ? ALLOCMAN_UT_DEV : ALLOCMAN_UT_KERNEL;
         if (utType == ALLOCMAN_UT_DEV &&
-            paddr >= linux_ram_paddr_base && paddr <= (linux_ram_paddr_base + (linux_ram_size - 1))) {
+            paddr >= ram_paddr_base && paddr <= (ram_paddr_base + (ram_size - 1))) {
             utType = ALLOCMAN_UT_DEV_MEM;
         }
         err = allocman_utspace_add_uts(allocman, 1, &path, &size, &paddr, utType);
@@ -471,8 +463,8 @@ static int vmm_init(void)
             cspacepath_t path;
             vka_cspace_make_path(vka, cap, &path);
             int utType = ALLOCMAN_UT_DEV;
-            if (paddr >= linux_ram_paddr_base &&
-                paddr <= (linux_ram_paddr_base + (linux_ram_size - 1))) {
+            if (paddr >= ram_paddr_base &&
+                paddr <= (ram_paddr_base + (ram_size - 1))) {
                 utType = ALLOCMAN_UT_DEV_MEM;
             }
             err = allocman_utspace_add_uts(allocman, 1, &path, &size, &paddr, utType);
@@ -636,7 +628,7 @@ static USED SECTION("_vmm_module") struct {} dummy_module;
 extern vmm_module_t __start__vmm_module[];
 extern vmm_module_t __stop__vmm_module[];
 
-int install_linux_devices(vm_t *vm)
+int install_vm_devices(vm_t *vm)
 {
     int err;
 
@@ -778,15 +770,15 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
         return -1;
     }
 
-    /* generate a memory node (linux_ram_base and linux_ram_size) */
-    err = fdt_generate_memory_node(gen_fdt, linux_ram_base, linux_ram_size);
+    /* generate a memory node (ram_base and ram_size) */
+    err = fdt_generate_memory_node(gen_fdt, ram_base, ram_size);
     if (err) {
         ZF_LOGE("Couldn't generate memory_node (%d)\n", err);
         return -1;
     }
 
-    /* generate a chosen node (linux_image_config.linux_bootcmdline, linux_stdout) */
-    err = fdt_generate_chosen_node(gen_fdt, linux_image_config.linux_stdout, linux_image_config.linux_bootcmdline,
+    /* generate a chosen node (vm_image_config.kernel_bootcmdline, kernel_stdout) */
+    err = fdt_generate_chosen_node(gen_fdt, kernel_stdout, kernel_bootcmdline,
                                    NUM_VCPUS);
     if (err) {
         ZF_LOGE("Couldn't generate chosen_node (%d)\n", err);
@@ -831,14 +823,14 @@ static int load_generated_dtb(vm_t *vm, uintptr_t paddr, void *addr, size_t size
     return 0;
 }
 
-static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, const char *initrd_name)
+static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, const char *initrd_name)
 {
     seL4_Word entry;
     seL4_Word dtb;
     int err;
 
     /* Install devices */
-    err = install_linux_devices(vm);
+    err = install_vm_devices(vm);
     if (err) {
         printf("Error: Failed to install Linux devices\n");
         return -1;
@@ -848,7 +840,7 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
 
     /* Load kernel */
     guest_kernel_image_t kernel_image_info;
-    err = vm_load_guest_kernel(vm, kernel_name, linux_ram_base, 0, &kernel_image_info);
+    err = vm_load_guest_kernel(vm, kernel_name, ram_base, 0, &kernel_image_info);
     entry = kernel_image_info.kernel_image.load_paddr;
     if (!entry || err) {
         return -1;
@@ -878,9 +870,8 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
         int dtb_fd = -1;
 
         /* No point checking the file server if the string is empty! */
-        if ((NULL != linux_image_config.dtb_base_name) &&
-            (linux_image_config.dtb_base_name[0] != '\0')) {
-            dtb_fd = open(linux_image_config.dtb_base_name, 0);
+        if ((NULL != dtb_base_name) && (dtb_base_name[0] != '\0')) {
+            dtb_fd = open(dtb_base_name, 0);
         }
 
         /* If dtb_base_name is in the file server, grab it and use it as a base */
@@ -925,17 +916,6 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
     }
 
     return 0;
-}
-
-void parse_camkes_linux_attributes(void)
-{
-    linux_ram_base = strtoul(linux_address_config.linux_ram_base, NULL, 0);
-    linux_ram_paddr_base = strtoul(linux_address_config.linux_ram_paddr_base, NULL, 0);
-    linux_ram_size = strtoul(linux_address_config.linux_ram_size, NULL, 0);
-    linux_ram_offset = strtoul(linux_address_config.linux_ram_offset, NULL, 0);
-    dtb_addr = strtoul(linux_address_config.dtb_addr, NULL, 0);
-    initrd_max_size = strtoul(linux_address_config.initrd_max_size, NULL, 0);
-    initrd_addr = strtoul(linux_address_config.initrd_addr, NULL, 0);
 }
 
 /* Async event handling registration implementation */
@@ -1097,8 +1077,6 @@ int main_continued(void)
     vm_t vm;
     int err;
 
-    parse_camkes_linux_attributes();
-
     /* setup for restart with a setjmp */
     while (setjmp(restart_jmp_buf) != 0) {
         err = vmm_process_reboot_callbacks(&vm, &reboot_hooks_list);
@@ -1188,7 +1166,7 @@ int main_continued(void)
     }
 
     /* Load system images */
-    err = load_linux(&vm, linux_image_config.linux_name, linux_image_config.dtb_name, linux_image_config.initrd_name);
+    err = load_vm(&vm, _kernel_name, _dtb_name, _initrd_name);
     if (err) {
         printf("Failed to load VM image\n");
         seL4_DebugHalt();
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -785,7 +785,7 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
         return -1;
     }
 
-    if (config_set(CONFIG_VM_INITRD_FILE)) {
+    if (provide_initrd) {
         err = fdt_append_chosen_node_with_initrd_info(gen_fdt, initrd_addr, initrd_size);
         if (err) {
             ZF_LOGE("Couldn't generate chosen_node_with_initrd_info (%d)\n", err);
@@ -829,12 +829,17 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
     seL4_Word dtb;
     int err;
 
+    vm->mem.map_one_to_one = map_one_to_one; /* Map memory 1:1 if configured to do so */
+
     /* Install devices */
     err = install_vm_devices(vm);
     if (err) {
         printf("Error: Failed to install Linux devices\n");
         return -1;
     }
+
+    vm->entry = entry_addr;
+    vm->mem.clean_cache = clean_cache;
 
     printf("Loading Kernel: \'%s\'\n", kernel_name);
 
@@ -848,7 +853,7 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
 
     /* Attempt to load initrd if provided */
     guest_image_t initrd_image;
-    if (config_set(CONFIG_VM_INITRD_FILE)) {
+    if (provide_initrd) {
         printf("Loading Initrd: \'%s\'\n", initrd_name);
         err = vm_load_guest_module(vm, initrd_name, initrd_addr, 0, &initrd_image);
         void *initrd = (void *)initrd_image.load_paddr;
@@ -857,7 +862,10 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
         }
     }
 
-    if (!config_set(CONFIG_VM_DTB_FILE)) {
+    ZF_LOGW_IF(provide_dtb && generate_dtb,
+               "provide_dtb and generate_dtb are both set. The provided dtb will NOT be loaded");
+
+    if (generate_dtb) {
         void *fdt_ori;
         void *gen_fdt = linux_gen_dtb_buf;
         int size_gen = PLAT_LINUX_DTB_SIZE;
@@ -896,7 +904,7 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
         vm_ram_touch(vm, dtb_addr, size_gen, load_generated_dtb, gen_fdt);
         printf("Loading Generated DTB\n");
         dtb = dtb_addr;
-    } else {
+    } else if (provide_dtb) {
         printf("Loading DTB: \'%s\'\n", dtb_name);
 
         /* Load device tree */
@@ -906,6 +914,9 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
         if (!dtb || err) {
             return -1;
         }
+    } else {
+        ZF_LOGW("%s not given a DTB - This may be appropriate for your guest, but it " \
+                "may also break things!", get_instance_name());
     }
 
     /* Set boot arguments */
--- a/components/VM_Arm/src/modules/init_ram.c
+++ b/components/VM_Arm/src/modules/init_ram.c
@@ -14,13 +14,7 @@
 
 void WEAK init_ram_module(vm_t *vm, void *cookie)
 {
-    int err;
-
-    if (config_set(CONFIG_PLAT_EXYNOS5) || config_set(CONFIG_PLAT_QEMU_ARM_VIRT) || config_set(CONFIG_PLAT_TX2)) {
-        err = vm_ram_register_at(vm, ram_base, ram_size, true);
-    } else {
-        err = vm_ram_register_at(vm, ram_base, ram_size, false);
-    }
+    int err = vm_ram_register_at(vm, ram_base, ram_size, vm->mem.map_one_to_one);
     assert(!err);
 }
 
--- a/components/VM_Arm/src/modules/init_ram.c
+++ b/components/VM_Arm/src/modules/init_ram.c
@@ -12,17 +12,14 @@
 #include <sel4vmmplatsupport/guest_memory_util.h>
 #include <vmlinux.h>
 
-extern unsigned long linux_ram_base;
-extern unsigned long linux_ram_size;
-
 void WEAK init_ram_module(vm_t *vm, void *cookie)
 {
     int err;
 
     if (config_set(CONFIG_PLAT_EXYNOS5) || config_set(CONFIG_PLAT_QEMU_ARM_VIRT) || config_set(CONFIG_PLAT_TX2)) {
-        err = vm_ram_register_at(vm, linux_ram_base, linux_ram_size, true);
+        err = vm_ram_register_at(vm, ram_base, ram_size, true);
     } else {
-        err = vm_ram_register_at(vm, linux_ram_base, linux_ram_size, false);
+        err = vm_ram_register_at(vm, ram_base, ram_size, false);
     }
     assert(!err);
 }
--- a/templates/seL4VMParameters.template.c
+++ b/templates/seL4VMParameters.template.c
@@ -17,9 +17,7 @@
 const unsigned long ram_base = /*? vm_address_config.get('ram_base') ?*/;
 const unsigned long ram_paddr_base = /*? vm_address_config.get('ram_paddr_base') ?*/;
 const unsigned long ram_size = /*? vm_address_config.get('ram_size') ?*/;
-const unsigned long ram_offset = /*? vm_address_config.get('ram_offset') ?*/;
 const unsigned long dtb_addr = /*? vm_address_config.get('dtb_addr') ?*/;
-const unsigned long initrd_max_size = /*? vm_address_config.get('initrd_max_size') ?*/;
 const unsigned long initrd_addr = /*? vm_address_config.get('initrd_addr') ?*/;
 
 /*- if vm_address_config.get('kernel_entry_addr') != '-1' -*/
--- a/templates/seL4VMParameters.template.c
+++ b/templates/seL4VMParameters.template.c
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2023, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <camkes.h>
+
+/*- set vm_address_config = configuration[me.name].get('vm_address_config') -*/
+/*- set vm_image_config = configuration[me.name].get('vm_image_config') -*/
+/*- set linux_address_config = configuration[me.name].get('linux_address_config') -*/
+/*- set linux_image_config = configuration[me.name].get('linux_image_config') -*/
+
+/*- if vm_address_config -*/
+
+const unsigned long ram_base = /*? vm_address_config.get('ram_base') ?*/;
+const unsigned long ram_paddr_base = /*? vm_address_config.get('ram_paddr_base') ?*/;
+const unsigned long ram_size = /*? vm_address_config.get('ram_size') ?*/;
+const unsigned long ram_offset = /*? vm_address_config.get('ram_offset') ?*/;
+const unsigned long dtb_addr = /*? vm_address_config.get('dtb_addr') ?*/;
+const unsigned long initrd_max_size = /*? vm_address_config.get('initrd_max_size') ?*/;
+const unsigned long initrd_addr = /*? vm_address_config.get('initrd_addr') ?*/;
+
+/*- else -*/
+
+#warning You are using the deprecated linux_address_config structure. Please use the vm_address_config structure instead
+
+const unsigned long ram_base = /*? linux_address_config.get('linux_ram_base') ?*/;
+const unsigned long ram_paddr_base = /*? linux_address_config.get('linux_ram_paddr_base') ?*/;
+const unsigned long ram_size = /*? linux_address_config.get('linux_ram_size') ?*/;
+const unsigned long ram_offset = /*? linux_address_config.get('linux_ram_offset') ?*/;
+const unsigned long dtb_addr = /*? linux_address_config.get('dtb_addr') ?*/;
+const unsigned long initrd_max_size = /*? linux_address_config.get('initrd_max_size') ?*/;
+const unsigned long initrd_addr = /*? linux_address_config.get('initrd_addr') ?*/;
+
+/*- endif -*/
+
+/*- if vm_image_config -*/
+
+const char *_kernel_name = "/*? vm_image_config.get('kernel_name') ?*/";
+const char *_dtb_name = "/*? vm_image_config.get('dtb_name', "") ?*/";
+const char *_initrd_name = "/*? vm_image_config.get('initrd_name', "") ?*/";
+const char *kernel_bootcmdline = "/*? vm_image_config.get('kernel_bootcmdline', "") ?*/";
+const char *kernel_stdout = "/*? vm_image_config.get('kernel_stdout', "") ?*/";
+const char *dtb_base_name = "/*? vm_image_config.get('dtb_base_name', "") ?*/";
+
+/*- else -*/
+
+#warning You are using the deprecated linux_image_config structure. Please use the vm_image_config structure instead
+
+const char *_kernel_name = "/*? linux_image_config.get('linux_name') ?*/";
+const char *_dtb_name = "/*? linux_image_config.get('dtb_name') ?*/";
+const char *_initrd_name = "/*? linux_image_config.get('initrd_name') ?*/";
+const char *kernel_bootcmdline = "/*? linux_image_config.get('linux_bootcmdline') ?*/";
+const char *kernel_stdout = "/*? linux_image_config.get('linux_stdout') ?*/";
+const char *dtb_base_name = "/*? linux_image_config.get('dtb_base_name') ?*/";
+
+/*- endif -*/
--- a/templates/seL4VMParameters.template.c
+++ b/templates/seL4VMParameters.template.c
@@ -5,6 +5,7 @@
  */
 
 #include <camkes.h>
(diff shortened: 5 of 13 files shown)


components/VM_Arm/src/main.c
```diff
@@ -882,7 +900,7 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
         vm_ram_touch(vm, dtb_addr, size_gen, load_generated_dtb, gen_fdt);
         printf("Loading Generated DTB");
         dtb = dtb_addr;
-    } else {
+    } else if (vm_image_config.provide_dtb) {
```
kent-mcleod, line 903: Does this mean that if generate_dtb or provide_dtb aren't set, then no dtb is loaded? Which would be fine for guests that don't require a dtb?

components/VM_Arm/src/modules/init_ram.c
```diff
@@ -17,14 +17,7 @@ extern unsigned long ram_size;
...
 {
-    int err;
-
-    if (config_set(CONFIG_PLAT_EXYNOS5) || config_set(CONFIG_PLAT_QEMU_ARM_VIRT) || config_set(CONFIG_PLAT_TX2)) {
-        err = vm_ram_register_at(vm, ram_base, ram_size, true);
-    } else {
-        err = vm_ram_register_at(vm, ram_base, ram_size, false);
-    }
-    assert(!err);
+    assert(!vm_ram_register_at(vm, ram_base, ram_size, vm->mem.unity));
```
kent-mcleod, line 20: Under Release builds, this assert would be removed by the compiler, and the call to vm_ram_register_at would be removed also.

components/VM_Arm/src/main.c
chrisguikema: Correct. I could add a ZF_LOG statement there to warn users. But specifically RTEMS doesn't need a DTB, which is why this was implemented.

components/VM_Arm/src/modules/init_ram.c
chrisguikema: Good to know. Will update.

components/VM_Arm/src/main.c
axel-h: We might want to catch mis-configuration cases. What happen in a LInux VM if none of the filed is set (by accident or dev error). Is the DTB a NULL then and and there is a way to see what is wrong from the logs?

components/VM_Arm/configurations/vm.h
```diff
@@ -68,22 +68,28 @@
...
         string dtb_name = "linux-dtb"; \
         string initrd_name = "linux-initrd"; \
-        string linux_bootcmdline = ""; \
-        string linux_stdout = ""; \
+        string kernel_bootcmdline = ""; \
+        string kernel_stdout = ""; \
         string dtb_base_name = ""; \
-    } linux_image_config; \
+        int provide_dtb = true; \
+        int generate_dtb = false; \
```
axel-h, line 88: It seems `generate_dtb` overwrites `provide_dtb`, which is not intuitive. To make the configuration more clear, we could merge these should into a singe `dtb_handling`. Then we might even drop `dtb_base_name`and use `dtb_name` here also.
- `VM_DTB_FIXED (0)`: provide DTB from `dtb_name`. If unset don't provide a DTB. This is the default.
- `VM_DTB_GENERATE (1)`: generate based on `dtb_name`. If unset the generated DTB is based on an empty DTB
- `VM_DTB_NONE (2)` 2: don't provide DTB. `dtb_name` is ignored
- ... other options reserved for future use?

components/VM_Arm/configurations/vm.h
```diff
@@ -68,22 +68,28 @@
...
+        int provide_initrd = true; \
+        int clean_cache = false; \
+        int map_unity = false; \
```
axel-h, line 91: I wonder if "one_to_one" is more intuitive as a name than "unity"?
```suggestion
        int map_one_to_one = false; \
```

components/VM_Arm/src/main.c
chrisguikema: Maybe add a ZF_LOGW?

components/VM_Arm/configurations/vm.h
chrisguikema: Done.

components/VM_Arm/configurations/vm.h
chrisguikema: I agree that DTB handling should be updated. This MR just updates the functionality to allow VMs to define their own per-VM settings. I like your suggestion, and I think it should be opened as an issue and made into its own MR.

components/VM_Arm/configurations/vm.h
```diff
@@ -147,4 +153,3 @@
 #define VM_VIRTUAL_SERIAL_CONFIGURATION_DEF(vm_ids...) \
     VM_VIRTUAL_SERIAL_GENERAL_CONFIGURATION_DEF() \
     __CALL(PER_VM_VIRTUAL_SERIAL_CONFIGURATION_DEF, vm_ids) \
-
```
axel-h, line 150: Please leave this line in.

chrisguikema: Done

components/VM_Arm/src/main.c
```diff
@@ -831,32 +832,48 @@ static int load_generated_dtb(vm_t *vm, uintptr_t paddr, void *addr, size_t size
...
     /* Install devices */
-    err = install_linux_devices(vm);
+    err = install_vm_devices(vm);
     if (err) {
         printf("Error: Failed to install Linux devices\n");
         return -1;
     }
 
+    /* If the entry_addr isn't set, just use the previous default logic */
+    if (!entry_addr) {
```
axel-h, line 851: - We should print a warning here to really encourage users to set this. It would be nice of we could remove any platform code here, so this must be parameter that i specified explicitly. That would make things more transparent. 
- Since we have guest OS agnostic now, can we be sure `entry_addr` can't be `0` also? This might be another argument to remove any smart default here that may cause strange side effects. It may break compatibility, but I wonder if this is a price to pay for a more well defined behavior without internal magic.

components/VM_Arm/src/main.c
```diff
@@ -927,15 +945,16 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
...
-    initrd_max_size = strtoul(linux_address_config.initrd_max_size, NULL, 0);
-    initrd_addr = strtoul(linux_address_config.initrd_addr, NULL, 0);
+    ram_base = strtoul(vm_address_config.ram_base, NULL, 0);
+    ram_paddr_base = strtoul(vm_address_config.ram_paddr_base, NULL, 0);
+    ram_size = strtoul(vm_address_config.ram_size, NULL, 0);
+    ram_offset = strtoul(vm_address_config.ram_offset, NULL, 0);
+    dtb_addr = strtoul(vm_address_config.dtb_addr, NULL, 0);
+    initrd_max_size = strtoul(vm_address_config.initrd_max_size, NULL, 0);
+    initrd_addr = strtoul(vm_address_config.initrd_addr, NULL, 0);
+    entry_addr = strtoul(vm_address_config.kernel_entry_addr, NULL, 0);
```
axel-h, line 957: I wonder if it's better to put the platform default magic here, because we can differentiate better
- parameter not set -> do platform default magic, but print a warning that this is deprecated and the entry address must be given explicitly
- parameter is not a string that contains an integer -> error
- parameter is 0 -> use this value
- parameter is an address -> use this value

components/VM_Arm/src/main.c
chrisguikema: I've added a ZF_LOGW if entry_addr is not set.

For your second point, it would technically be possible to have RTEMS boot at 0, but for Linux, there's always an offset.

components/VM_Arm/src/main.c
chrisguikema: I added a function that parses the entry_addr parameter. Default is set to `-1`, which won't trigger an error on build, but will print a warning that it was calculated.

components/VM_Arm/src/main.c
```diff
@@ -927,15 +938,38 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
     return 0;
 }
 
-void parse_camkes_linux_attributes(void)
+unsigned int parse_camkes_vm_entry_addr(void)
```
axel-h, line 941:
```suggestion
static unsigned int parse_camkes_vm_entry_addr(void)
```

components/VM_Arm/src/main.c
```diff
@@ -927,15 +938,38 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
...
+            addr = base + 0x8000;
+        }
+        ZF_LOGW("Warning - %s.vm_address_config.kernel_entry_addr is not set. The entry address \
+                has been set to 0x%lx", get_instance_name(), addr);
+    }
+
+    return addr;
+}
+
+void parse_camkes_vm_attributes(void)
```
axel-h, line 963: I thing this is also just an internal helper function for this file, so let's not expose it.
```suggestion
static void parse_camkes_vm_attributes(void)
```

components/VM_Arm/src/main.c
axel-h: Thanks, that looks really nice now.

components/VM_Arm/src/main.c
```diff
@@ -927,15 +938,38 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
     return 0;
 }
 
-void parse_camkes_linux_attributes(void)
+unsigned int parse_camkes_vm_entry_addr(void)
+{
+    unsigned long addr = strtoul(vm_address_config.kernel_entry_addr, NULL, 0);
```
axel-h, line 943: We have to set ENTRY_ADDR_NOT_SET here, right?
```suggestion
    unsigned long addr = strtoul(vm_address_config.kernel_entry_addr, NULL, ENTRY_ADDR_NOT_SET);
```

axel-h: No, ignore that.

chrisguikema: The third argument of `strtoul` isn't the default value, its a base.

https://www.tutorialspoint.com/c_standard_library/c_function_strtoul.htm

The default being -1 is set here: https://github.com/seL4/camkes-vm/pull/48/files#diff-54f626392a451bea7096290caa1f7d5793d6909bf49da22e0ff19a52c659ca92R78

components/VM_Arm/src/main.c
chrisguikema: Good point. Will make it static.

components/VM_Arm/configurations/vm.h
```diff
@@ -68,22 +68,28 @@
...
-        string linux_ram_offset; \
+        string ram_base; \
+        string ram_paddr_base; \
+        string ram_size; \
+        string ram_offset; \
         string dtb_addr; \
         string initrd_max_size; \
         string initrd_addr; \
-    } linux_address_config; \
+        string kernel_entry_addr = "-1"; \
```
axel-h, line 78: We can use `ENTRY_ADDR_NOT_SET`here:
```suggestion
        string kernel_entry_addr = VAR_STRINGIZE(ENTRY_ADDR_NOT_SET); \
```

components/VM_Arm/src/main.c
chrisguikema: ZF_LOGW added to warn users of a potentially breaking state.

components/VM_Arm/configurations/vm.h
axel-h: Do you have a dependency on the two variables already and want to get them in? Otherwise, I'd argue, that instead of introducing these two here now, just introduce `int dtb_handling` with the 3 values I suggested above. Then we don't introduce one thing just to remove it again later.

components/VM_Arm/configurations/vm.h
chrisguikema: When I do that, I get:

```
kernel_entry_addr = "ENTRY_ADDR_NOT_SET",
```

`strtoul` defaults to zero, so  the logic wouldn't apply, since technically a guest could have an entry address of zero. We'd have to pass that flag through via the `cpp_flags`

```
set(cpp_flags "-DKERNELARMPLATFORM_EXYNOS5410")
```

components/VM_Arm/configurations/vm.h
chrisguikema: I don't have a dependency on it, my concern is delaying the PR. 

This PR basically updates the functionality that already exists, where your suggested changes expand the functionality to something beyond the scope of this PR. If I can find time, I can implement this method, but I'd like to make it a separate PR.

chrisguikema: I added a warning print in the case where both `provide_dtb` and `generate_dtb` are set that warns the user that the provided dtb will not be loaded.

components/VM_Arm/configurations/vm.h
axel-h: Hmm.... I though the double macro layers in `VAR_STRINGIZE` are supposed to solve exactly this. Then let's remove the define `ENTRY_ADDR_NOT_SET` and use `-1` in both cases. We can clean this up later then

chrisguikema: That's how I currently have it. `ENTRY_ADDR_NOT_SET` is defined as `-1`, which makes the C code more readable.

axel-h (approved):
This looks good to get merged now for me.

components/VM_Arm/configurations/vm.h
axel-h: Fair point, let's keep it then.

chrisguikema:
Thanks for the review, Axel. Appreciate it!

axel-h:
> Thanks for the review, Axel. Appreciate it!

The stuff in these PR is really useful and having all these config option do avoid some hacks on our side, too. Also, now that I have looked into this again, this explains some things I have been really wondering about...

components/VM_Arm/configurations/vm.h
```diff
@@ -68,22 +68,28 @@
     attribute int num_extra_frame_caps; \
     attribute int extra_frame_map_address; \
     attribute { \
-        string linux_ram_base; \
-        string linux_ram_paddr_base; \
-        string linux_ram_size; \
-        string linux_ram_offset; \
+        string ram_base; \
```
axel-h, line 71: Since we are doing a breaking change here anyway, I wonder if we should also change `string`to `uint64_t` here. There seems no point why these are string, except that in the past this was necessary because `uint64_t` did not exist.

components/VM_Arm/src/main.c
```diff
@@ -856,7 +864,7 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
 
     /* Attempt to load initrd if provided */
     guest_image_t initrd_image;
-    if (config_set(CONFIG_VM_INITRD_FILE)) {
+    if (vm_image_config.provide_initrd) {
```
hlyytine, line 867: How about just `if (strlen(initrd_name))`? Then you could set `initrd_name` to empty string in CAmkES file to not load any.

components/VM_Arm/src/main.c
```diff
@@ -865,7 +873,10 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
         }
     }
 
-    if (!config_set(CONFIG_VM_DTB_FILE)) {
+    ZF_LOGW_IF(vm_image_config.provide_dtb && vm_image_config.generate_dtb,
```
hlyytine, line 876: How about getting rid of both these booleans and reserving some magic string for autogeneration. For example, setting `dtb_name` to `<generate>` is not a likely filename since it would create havoc in build systems due to `<` and `>` being shell redirections. The second case is leaving `dtb_name` empty and no DTB is loaded. Any other string should be considered a filename.

components/VM_Arm/src/main.c
```diff
@@ -927,6 +941,28 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
     return 0;
 }
 
+static unsigned long parse_camkes_vm_entry_addr(void)
+{
+    unsigned long addr = strtoul(vm_address_config.kernel_entry_addr, NULL, 0);
```
hlyytine, line 946: Perhaps `uintptr_t` is more appropriate for these?

components/VM_Arm/src/main.c
```diff
@@ -927,6 +941,28 @@ static int load_vm(vm_t *vm, const char *kernel_name, const char *dtb_name, cons
...
+    unsigned long base = strtoul(vm_address_config.ram_base, NULL, 0);
+
+    /* Calculate an entry address if not explicitly set */
+    if (ENTRY_ADDR_NOT_SET == addr) {
+        if (config_set(CONFIG_PLAT_TX1) || config_set(CONFIG_PLAT_TX2) ||
+            config_set(CONFIG_PLAT_QEMU_ARM_VIRT) || config_set(CONFIG_PLAT_ODROIDC2) ||
+            config_set(CONFIG_PLAT_ZYNQMP)) {
+            /* This is likely an aarch64/aarch32 linux difference */
```
hlyytine, line 954: We think so too, so how about using `CONFIG_ARCH_AARCH64` instead?

components/VM_Arm/src/main.c
hlyytine, line 946: What if `kernel_entry_addr` was an empty string by default and you used `if (!strlen(vm_address_config.kernel_entry_addr))` to branch to computing entry address based on architecture? No need to have yet another `#define`.

components/VM_Arm/configurations/vm.h
```diff
@@ -68,22 +68,28 @@
...
-        string linux_stdout = ""; \
+        string kernel_bootcmdline = ""; \
+        string kernel_stdout = ""; \
         string dtb_base_name = ""; \
-    } linux_image_config; \
+        int provide_dtb = true; \
+        int generate_dtb = false; \
+        int provide_initrd = true; \
+        int clean_cache = false; \
+        int map_one_to_one = false; \
```
hlyytine, line 91: Just a suggestion, could `unity_mapping` be more descriptive? Or `stage2_unity_mapping`?

axel-h: It's called `map_one_to_one`, isn't this clear enough?

components/VM_Arm/src/main.c
axel-h: Yes, this seems a better choice.

components/VM_Arm/src/main.c
axel-h: Yes, that would be sufficient to keep the config options lighter. Shoul we put this into https://github.com/seL4/camkes-vm/blob/296524002d908c475f364fcfbcbfa41496a5d2ee/components/VM_Arm/configurations/vm.h#L82 then:
```
string initrd_name = ""; /* no inird by default */ \
``` 
then as default in https://github.com/seL4/camkes-

components/VM_Arm/src/main.c
axel-h: Actually, there are multiple options, see also my comment https://github.com/seL4/camkes-vm/pull/48#discussion_r1007410679, so using your proposal we'd have:
- don't provide DTB. Could be `dtb_name = ""`
- provide a fixed TB from `dtb_name`. Do not modify it. Could be `dtb_name = "some.dtb"`
- generate as DTB from scratch. Could be `dtb_name = "<generate>"`
- generate as DTB based on a give DTB. Should this be handled via `dtb_name = "<generate>some.dtb"`

components/VM_Arm/configurations/vm.h
hlyytine: I think it's not immediately clear this is about MMU.

components/VM_Arm/src/main.c
hlyytine: Actually we have that type fix commit in our forks already, I can make a PR soon.

components/VM_Arm/src/main.c
axel-h: So, let's take the chance and drop it - unless anybody (SIG-VMM?) has a strong interest to keep this?

components/VM_Arm/src/main.c
hlyytine: Here: seL4/camkes-vm#58

components/VM_Arm/src/main.c
hlyytine: How about:

- Don't provide DTB: `dtb_name = ""`
- Provide fixed DTB: `dtb_name = "some.dtb"`
- Generate from scratch: `dtb_name = "<"`
- Generate from given DTB: `dtb_name = "<some.dtb"`

Note dropped `>` at the end. Now you can just check if first character is `<`, if yes, use the remaining string (which might be empty) as filename. No need for error prone string copies. Any other character than `<` would be fine to me, as long as it is unlikely to appear in real world filenames.

axel-h: That looks nice actually. We just have to clearly document this somewhere then also.

components/VM_Arm/src/main.c
hlyytine: Sorry @chrisguikema for stealing the show, but let me show how I think this per-VM initrd is handled best: seL4/camkes-vm#59.

components/VM_Arm/configurations/vm.h
chrisguikema: I actually had it as `map_unity` originally, but it was requested that it changed to `map_one_to_one` by @wom-bat.

I feel like its fairly clear, but maybe some documentation about how to use the setting would be beneficial?

components/VM_Arm/src/main.c
hlyytine: Figured out we still have 3 special cases for guest's data blobs: kernel, DTB and initrd. I got an idea how to write that in a more generic way and remove lots of code from VM. Please give me a day or two so I can show my suggestion.

chrisguikema: I appreciate the review, Hannu, and I also appreciate that you're creating new MRs that address your comments

components/VM_Arm/src/main.c
chrisguikema: I'll add a commit dropping this for CONFIG_ARCH_AARCH64

components/VM_Arm/src/main.c
chrisguikema: That was the behavior I had originally. After discussions with @axel-h we implemented it in a way that forces the user to define the value.

However, if community feedback wants implemented in a different manner, I can implement in that way.

Thoughts @kent-mcleod @wom-bat @lsf37?

components/VM_Arm/configurations/vm.h
kent-mcleod: I think it was @axel-h who requested the one-to-one (https://github.com/seL4/camkes-vm/pull/48#discussion_r1035041892)?

Maybe something like `ram_stage2_direct_translate`? Because it's only being applied to the ram device and the effect is that the stage2 translation is emulating direct-mapping with it's translations...

components/VM_Arm/configurations/vm.h
```diff
@@ -68,22 +68,28 @@
...
-        string linux_ram_paddr_base; \
-        string linux_ram_size; \
-        string linux_ram_offset; \
+        string ram_base; \
+        string ram_paddr_base; \
+        string ram_size; \
+        string ram_offset; \
         string dtb_addr; \
         string initrd_max_size; \
         string initrd_addr; \
```
kent-mcleod, line 77: It seems like there's an easy way to implement these configuration improvements without breaking any existing project CAmkES configurations. (There's lots of projects, examples, tutorials, and old emails sitting around that use these configs). If we defined vm_address_config and vm_image_config as new attributes while leaving the existing attribute maps around then new projects could choose to use the new config types and old projects and docs will continue to work.

The VM component can then require that if any of the new settings are set, the old settings must be unset or something.

components/VM_Arm/configurations/vm.h
chrisguikema: I'm okay with whatever the community decides is best. 

Either way, I'll push up a commit that documents what the flag does.

Just for a simple poll, what do people like best?

- map_unity
- map_one_to_one
- ram_stage2_direct_translate
- stage2_unity_mapping
- ram_direct_translate

Personally, I think referencing the stage 2 mmu gets a bit too complex. How many people even know about how the MMU handles stage 2 mappings?

components/VM_Arm/configurations/vm.h
chrisguikema: Good idea. I'll add that to my "dereference linux" commit. I'll put some warning about using a deprecated option.

axel-h: If we keep compatibility, then the new values should really move from `string` to `uint64_t` for all new address and offset attributes. There seem no reason to keep them as strings.

components/VM_Arm/configurations/vm.h
axel-h: I think it's somehow implicit that this affects the stage 2 translations only. So `ram_direct_translate` or `map_ram_one_to_one`. I have not seen the term "unity" getting used for such translation, so I find this a bit confusing. 
However, I also wonder if a direct mapping is usual intention anyway when the addresses match. So can't we make this the default then? A flag that enables a relaxed mapping might be something to add then - if there is really a need for this.

components/VM_Arm/configurations/vm.h
chrisguikema: I'm using templates to create a file containing the necessary values. It can pull from `vm_image_config` or `linux_image_config`. So its fine to leave them as strings, because the end result is this:

```
#warning You are using the deprecated linux_address_config structure. Please use the vm_address_config structure instead
#warning You are using the deprecated linux_image_config structure. Please use the vm_image_config structure instead

unsigned long ram_base = 0x40000000;
unsigned long ram_paddr_base = 0x40000000;
unsigned long ram_size = 0x20000000;
unsigned long ram_offset = 0x00000000;
unsigned long dtb_addr = 0x4F000000;
unsigned long initrd_max_size = 0x1900000;
unsigned long initrd_addr = 0x4D700000;
```

components/VM_Arm/src/main.c
```diff
@@ -927,17 +951,6 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
     return 0;
 }
 
-void parse_camkes_linux_attributes(void)
-{
-    linux_ram_base = strtoul(linux_address_config.linux_ram_base, NULL, 0);
```
axel-h, line 932: It we can avoid calling `strtoul()` at runtime, that's seem really nice.

components/VM_Arm/src/modules/init_ram.c
```diff
@@ -12,18 +12,12 @@
 #include <sel4vmmplatsupport/guest_memory_util.h>
 #include <vmlinux.h>
 
-extern unsigned long linux_ram_base;
-extern unsigned long linux_ram_size;
+extern unsigned long ram_base;
```
axel-h, line 15: Seem we could get this from a template header file then nicely?

chrisguikema: Depends on how we want to use the templates.

I could create wrapper functions that simply return a static variable contained in a template, or we can keep the `extern` method. Both work fine

kent-mcleod:
> Do you feel like the template solution works here?

Yea. Internal templates as configuration is an intended use-case of per-component templates. 

> Of course there will still be runtime issues with the user hasn't set the dtb and initrd flags properly, but that's not something we can control.

Would `provide_initrd || CONFIG_SET(CONFIG_VM_INITRD_FILE)` and ` provide_dtb || CONFIG_SET(CONFIG_VM_DTB_FILE)` work for this? Projects that currently set these config options would then keep the same behavior while updated projects can leave the options off and use the camkes properties?

components/VM_Arm/configurations/vm.h
kent-mcleod: Thanks for addressing this request so late in the review cycle @chrisguikema !

chrisguikema:
> Would provide_initrd || CONFIG_SET(CONFIG_VM_INITRD_FILE) and  provide_dtb || CONFIG_SET(CONFIG_VM_DTB_FILE) work for this?

Can we mix in @hlyytine's solution here? `provide_initrd` and `provide_dtb` would be the default behavior if explicitly set by the user. If they aren't set, then we set them to the value of the `CONFIG_SET` parameters while providing a warning that the user should update their configuration files.

components/VM_Arm/configurations/vm.h
chrisguikema: No worries, Kent. I think this feature would help out a lot of people, and it gets DornerWorks one step closer to having all our stuff open-sourced, so I'm happy to make the changes necessary to get this merged.

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2023, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <camkes.h>
+
+/*- if configuration[me.name].get('vm_address_config') -*/
```
kent-mcleod, line 9: It's curious that this isn't always set because of the default attribute values.

kent-mcleod:
> Can we mix in @hlyytine's solution here? provide_initrd and provide_dtb would be the default behavior if explicitly set by the user. If they aren't set, then we set them to the value of the CONFIG_SET parameters while providing a warning that the user should update their configuration files.

I thought that @hlyytine's solution involved a breaking change that would generate a build time error that someone would then be able to resolve?

I think the difference between what you're suggesting is whether setting provide_initrd to false overrides setting the CONFIG option to true? I'm fine with that approach too.

templates/seL4VMParameters.template.c
chrisguikema: Yeah I wondered about that. The camkes structure was defined in the header file, but the actual struct wasn't in the source.

chrisguikema:
I think the issue is specifically w/ DTB Generation.

The previous logic was:

```
if(!CONFIG_SET(CONFIG_VM_DTB_FILE)) {
 // generate dtb
} else {
 // use provided dtb
}
```

If we use that logic with the new flags, we would always generate a device tree if the variables aren't set.

chrisguikema:
There's also the issue that the vm_image_config flags overwrote some "per-platform" settings, like the cache_cleaning and mapping one to one were previously done via a CONFIG_PLAT check.

components/VM_Arm/src/main.c
chrisguikema: Agreed!

hlyytine:
I have pondered this the whole day and I propose the following: Let's add a field that describes the configuration file format:

`vm0.configuration_format_version = 1;`

By default (in `VM_Arm/configurations/vm.h`) this would be 0, meaning legacy code. And we would not change the semantics in those settings at all. But we would bug the developers (build-time nagging) and the end users (run-time nagging) about using deprecated configuration format version. Jinja2 templates and CPP are the tools here.

For configuration format version greater than zero we are free to implement entirely new version of `load_linux()`.  And having it as an integer gives us tools to harass (but not make their lives impossible) developers whenever we need to switch configuration format version 2 and so on.

And we definitely can say no no if somebody claims to use new configuration format version, yet they try to nudge the legacy options from their default values (including CMake options).

I have already written some code (C, Jinja2) which allows CAmkES VM file have arbitrary number of images to be loaded into guest RAM, be they raw or ELF files. I would like to ask a day or two so I can show you concrete code.

kent-mcleod:
> If we use that logic with the new flags, we would always generate a device tree if the variables aren't set.

Ok, I think I follow your point.

chrisguikema:
@kent-mcleod 

```
#ifdef CONFIG_VM_INITRD_FILE
#warning VmInitRdFile is a deprecated setting. Please remove and use vm_image_config.provide_initrd
int provide_initrd = 1;
#else
int provide_initrd = 0;
#endif

#ifdef CONFIG_VM_DTB_FILE
#warning VmDtbFile is a deprecated setting. Please remove and use vm_image_config.provide_dtb
int provide_dtb = 1;
int generate_dtb = 0;
#else
int provide_dtb = 0;
int generate_dtb = 1;
#endif

#if defined(CONFIG_PLAT_EXYNOS5) || defined(CONFIG_PLAT_QEMU_ARM_VIRT) || defined(CONFIG_PLAT_TX2)
int map_one_to_one = 1;
#else
int map_one_to_one = 0;
#endif

#if defined(CONFIG_PLAT_TX1) || defined(CONFIG_PLAT_TX2)
int clean_cache = 1;
#else
int clean_cache = 0;
#endif
```

Here is my logic. This sets the flags based on a combination of the CMake variables and previous behavior. I tested with `vm_minimal` on `master` and `multiple_os` branches and it worked for both.

kent-mcleod:
> `vm0.configuration_format_version = 1;`

I agree that something like this would be a good approach. Is it something that we could implement for the next iteration?

chrisguikema:
> Is it something that we could implement for the next iteration?

I can put it together. Do you want it as a part of this round of PRs?

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,104 @@
...
+
+unsigned long ram_base = /*? configuration[me.name].get('vm_address_config').get('ram_base') ?*/;
+unsigned long ram_paddr_base = /*? configuration[me.name].get('vm_address_config').get('ram_paddr_base') ?*/;
+unsigned long ram_size = /*? configuration[me.name].get('vm_address_config').get('ram_size') ?*/;
+unsigned long ram_offset = /*? configuration[me.name].get('vm_address_config').get('ram_offset') ?*/;
+unsigned long dtb_addr = /*? configuration[me.name].get('vm_address_config').get('dtb_addr') ?*/;
+unsigned long initrd_max_size = /*? configuration[me.name].get('vm_address_config').get('initrd_max_size') ?*/;
+unsigned long initrd_addr = /*? configuration[me.name].get('vm_address_config').get('initrd_addr') ?*/;
+
+/*- if configuration[me.name].get('vm_address_config').get('kernel_entry_addr') == -1 -*/
```
axel-h, line 20:
```suggestion
/*- if configuration[me.name].get('vm_address_config').get('kernel_entry_addr') != -1 -*/
```

kent-mcleod:
Or even better would be if we could use a thread on the new VMM SIG discourse to organize a sequence of issues to collectively resolve in order.  For example, SMC caps, x86 64bit guests and SMMUv2, GICv32 are all open prs that I'd prefer us to spend time on resolving and getting merged. This may be easier if we had a way to organize around an order of features we can work towards completion.

templates/seL4VMParameters.template.c
chrisguikema: Nice catch. I also had to check for `'-1'`  since its technically a string

chrisguikema:
I agree. Its always good to have a discussion, come up with the set of features we'd like addressed, and then match a PR to the discussion. 

With regards to this feature, I think its in a good place for merging. This latest rounds of comments have the feature implemented in such a way that existing applications won't break, which was a big concern of yours, Kent.

I believe the only outstanding issue is what to name the `map_one_to_one` flag, which is hopefully a quick discussion point before the feature can be merged.

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright 2023, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <camkes.h>
+#include <arm_vm/gen_config.h>
+
+/*- if configuration[me.name].get('vm_address_config') -*/
```
axel-h, line 10: Can we cache `configuration[me.name].get('vm_address_config')` in a script variable, so this becomes a bit easier to red and less redundant?

components/VM_Arm/configurations/vm.h
hlyytine: I can live with `map_one_to_one`, not losing my sleep.

templates/seL4VMParameters.template.c
chrisguikema: Sure thing.

components/VM_Arm/configurations/vm.h
chrisguikema: Thanks, Hannu.

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,115 @@
...
+/*- set linux_image_config = configuration[me.name].get('linux_image_config') -*/
+
+/*- if vm_address_config -*/
+
+unsigned long ram_base = /*? vm_address_config.get('ram_base') ?*/;
+unsigned long ram_paddr_base = /*? vm_address_config.get('ram_paddr_base') ?*/;
+unsigned long ram_size = /*? vm_address_config.get('ram_size') ?*/;
+unsigned long ram_offset = /*? vm_address_config.get('ram_offset') ?*/;
+unsigned long dtb_addr = /*? vm_address_config.get('dtb_addr') ?*/;
+unsigned long initrd_max_size = /*? vm_address_config.get('initrd_max_size') ?*/;
```
axel-h, line 22: Do we want to drop this, because it is not used anywhere? It could be added if there is a use case.

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,115 @@
...
+/*- set vm_image_config = configuration[me.name].get('vm_image_config') -*/
+/*- set linux_address_config = configuration[me.name].get('linux_address_config') -*/
+/*- set linux_image_config = configuration[me.name].get('linux_image_config') -*/
+
+/*- if vm_address_config -*/
+
+unsigned long ram_base = /*? vm_address_config.get('ram_base') ?*/;
+unsigned long ram_paddr_base = /*? vm_address_config.get('ram_paddr_base') ?*/;
+unsigned long ram_size = /*? vm_address_config.get('ram_size') ?*/;
+unsigned long ram_offset = /*? vm_address_config.get('ram_offset') ?*/;
```
axel-h, line 20: Is this used anywhere?

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright 2023, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <camkes.h>
+#include <arm_vm/gen_config.h>
+
```
axel-h, line 9: Please create a `seL4VMParameters.template.h` with the interface, so this can be included where needed. This avoids redundant manual definitions scattered over various c files.

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,115 @@
...
+/*- set vm_address_config = configuration[me.name].get('vm_address_config') -*/
+/*- set vm_image_config = configuration[me.name].get('vm_image_config') -*/
+/*- set linux_address_config = configuration[me.name].get('linux_address_config') -*/
+/*- set linux_image_config = configuration[me.name].get('linux_image_config') -*/
+
+/*- if vm_address_config -*/
+
+unsigned long ram_base = /*? vm_address_config.get('ram_base') ?*/;
```
axel-h, line 17: Actually, these are all constants and they could be put into a ro-segement.
```suggestion
const unsigned long ram_base = /*? vm_address_config.get('ram_base') ?*/;
```

templates/seL4VMParameters.template.c
chrisguikema: I'm not sure. If its unused, it should be removed as part of a new PR.

axel-h: I can't find a reference. Since we are creating a new block here, I'd leave everything out that has no clear use case, as there seem no gain just copying legacy. So, does anybody know where this is used?

templates/seL4VMParameters.template.c
chrisguikema: Done.

templates/seL4VMParameters.template.c
chrisguikema: Shouldn't be needed since all the variables are constants now.

axel-h: I consider it a good practice if a C file has a header with the interface. So everybody must includes the header instead of pulling in definitions directly in demand. It's better controlled then in the long run, and there is less chance different definitions exists in the code base that can break unnoticed. We have two C files already that copy the same definitions.

templates/seL4VMParameters.template.c
```diff
@@ -0,0 +1,115 @@
...
+unsigned long ram_paddr_base = /*? vm_address_config.get('ram_paddr_base') ?*/;
+unsigned long ram_size = /*? vm_address_config.get('ram_size') ?*/;
+unsigned long ram_offset = /*? vm_address_config.get('ram_offset') ?*/;
+unsigned long dtb_addr = /*? vm_address_config.get('dtb_addr') ?*/;
+unsigned long initrd_max_size = /*? vm_address_config.get('initrd_max_size') ?*/;
+unsigned long initrd_addr = /*? vm_address_config.get('initrd_addr') ?*/;
+
+/*- if vm_address_config.get('kernel_entry_addr') != '-1' -*/
```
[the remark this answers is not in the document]
chrisguikema: It needs to be set or else it will cause an KeyError during the `init-build.sh` script. So keeping the -1 default value allows for the automatic calculation to function if its not set by the user.

templates/seL4VMParameters.template.c
chrisguikema: Again, I'm not opposed to removing it, it just feels like that is a feature that deserves its own PR as its not related to the overall topic of this PR.

templates/seL4VMParameters.template.c
chrisguikema: Im confused. Do you want me to implement functions like this in the C source, and then replace everywhere the global variable is used in main.c?

```
int get_ram_base(void) {
    return ram_base;
}
```

templates/seL4VMParameters.template.c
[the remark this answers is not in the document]
axel-h: No sure I follow, wan't the point that we want the user to set this explicitly in the configuration, so it's more transparent what happens and there is less magic at the bottom? We could have an explicit linux template that provides defaults for Linux VMs, so users can use this one. Or a flag that says it's a linux kernel, which then enabled these well known defaults and maybe more convenince features. Also, it could be addressed in a meta VM creation system, that has more insight about the actual VMs, like the "vm composer" you have.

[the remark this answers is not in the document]
chrisguikema: I'm saying that every member of the vm_address_config struct _must_ have a value or the init-build.sh script will fail. We want to encourage the users to explicitly set the value, so when it is not set, CAmkES calculates a default, based off linux's offset, and the following warning will appear:

```
[258/305] Building C object CMakeFiles/vm0.instance.bin.dir/vm0/seL4VMParameters.template.c.obj
vm0/seL4VMParameters.template.c:27:2: warning: #warning Entry address has been calculated. Please use vm_address_config.kernel_entry_addr [-Wcpp]
   27 | #warning Entry address has been calculated. Please use vm_address_config.kernel_entry_addr
      |  ^~~~~~~
```

We need the `-1` default value because if the user didn't set the value, the configuration wouldn't even be able to compile. The `#warning` is what encourages the user to set the value.

templates/seL4VMParameters.template.c
axel-h: It just a standard C header file. So for any 
```
const unsigned long ram_size = /*? vm_address_config.get('ram_size') ?*/;
```
there is this in the header:
```
extern const unsigned long ram_size;
```
And whoever need something from the C file (in our case `main.c` and `init_ram.c`) has to include the header instead of directly cherry picking the extern declaration. It seem a bit overkill here, but it's worth it for the long term maintainability of the template.

chrisguikema: Makes sense. That's a good change. Done.

templates/seL4VMParameters.template.h
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2023, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
```
axel-h, line 6: Please add an include guard. Seem we don't need any special headers here as long as we just use plain C integers.

```suggestion
#pragma once
```

templates/seL4VMParameters.template.c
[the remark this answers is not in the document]
axel-h: Ah, I see. So I wonder, should the complete macro `VM_INIT_DEF()` be deprecated then? It is creating a `vm_address_config` and a `linux_address_config` block now, which seems a bit odd. Is it somehow possible to not have it create a `vm_address_config` block and then print a warning if that it's outdated? And there is another new macro (maybe `VM_GENERIC_INIT_DEF()`) to be used that creates only a `vm_address_config` now. And if this new macro is used, the all fields must be set by the user?

templates/seL4VMParameters.template.c
axel-h: True, it's just that we've moved now from changing name to creating a new config block. And we keep the old one for compatibility for a while. So, for the new block we can start small and just pass the test and use cases we have. If somebody runs into an issue during migration later, we can always extend things easily, it's a small PR and then also a clear pointer where this is used. It's much harder to remove something that has a blurry usage already.

components/VM_Arm/docs/vm_memory_configuration.md
```diff
@@ -0,0 +1,83 @@
...
+};
+
+vm0.simple_untyped24_pool = 33;
+
+```
+
+This configuration lets VM0 see 512MB at 0x40000000. When running as a hypervisor, seL4 uses stage 2
+mappings to map an Intermediate Physical Address (IPA) to a Physical Address (PA). This means that
+0x40000000 is the IPA that the guest sees, and the associated PA could be any available page in the
+system. The pages backing the guest mappings are provided by the `simple_untyped24_pool`.
```
axel-h, line 31: A note here would be good, why there have to be 33 and not just 32 memory blocks in the pool.

templates/seL4VMParameters.template.c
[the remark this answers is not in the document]
chrisguikema: I think if we have a version, like Hannu was suggesting, we can only include the necessary attribute. So for now this is okay, but we can make that part of the next discussion.

templates/seL4VMParameters.template.h
chrisguikema: Will do.

components/VM_Arm/docs/vm_memory_configuration.md
chrisguikema: Done.

templates/seL4VMParameters.template.c
chrisguikema: @kent-mcleod do you think that removing the initrd_max_size and ram_offset should be a part of this commit?

components/VM_Arm/configurations/vm.h
```diff
@@ -68,13 +68,36 @@
     attribute int num_extra_frame_caps; \
     attribute int extra_frame_map_address; \
     attribute { \
-        string linux_ram_base; \
```
axel-h, line 71: Since we keep the `linux_address_config` block, could you change this commit to not delete this here, but keep it and add the new `vm_address_config` block below? That makes the actual chances a bit easier to track.

templates/seL4VMParameters.template.c
kent-mcleod: ram_offset's dependency was removed in https://github.com/seL4/camkes-vm-examples/commit/da44d5b3c6ca9dece94163dc170334cef1ebabf3.  It seems like that would indicate that ram_offset was only ever set to 0.  I can't find where initrd_max_size was ever used. Now that we've looked it up, it's probably worth removing them from the new config variation.

components/VM_Arm/configurations/vm.h
chrisguikema: The issue was a couple of `=""` values that made the change harder to track. Its fixed now.

templates/seL4VMParameters.template.c
chrisguikema: Alright, I've removed them from the new config variation, and I also remove some extra macros that were in the vmlinux.h files.

templates/seL4VMParameters.template.c
chrisguikema: Drop in the new vm_address_config block. It has been left in the linux_address_config block for any backwards compatibility.

kent-mcleod (approved):

kent-mcleod:
@chrisguikema Can you let me know if you're ready for these 3 PRs to be merged.

chrisguikema:
Yeah I'm all good on my end!

PR #50, opened 2022-10-26
vm_arm: clarify on-demand device prints

chrisguikema:
Previously, the following prints would appear for any RAM-backed device install:

alloc_vm_device_cap@main.c:1011 Grabbing the entire cap for device memory 
alloc_vm_device_cap@main.c:1014 Failed to grab the entire cap

To an unexperienced user, this could cause confusion as to what is happening in the VMM. This commit reduces the verbosity of some prints, and adds prints to clarify the region mapped, and whether it is RAM backed or device backed.

Signed-off-by: Chris Guikema <chris.guikema@dornerworks.com>

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -887,7 +887,7 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
         }
         vm_ram_mark_allocated(vm, dtb_addr, size_gen);
         vm_ram_touch(vm, dtb_addr, size_gen, load_generated_dtb, gen_fdt);
-        printf("Loading Generated DTB");
+        printf("Loading Generated DTB\n");
         dtb = dtb_addr;
     } else {
         printf("Loading DTB: \'%s\'\n", dtb_name);
@@ -992,10 +992,10 @@ static int alloc_vm_device_cap(uintptr_t addr, vm_t *vm, vm_frame_t *frame_resul
     err = vka_utspace_alloc_at(vm->vka, &frame, kobject_get_type(KOBJECT_FRAME, seL4_PageBits), seL4_PageBits, addr,
                                &cookie);
     if (err) {
-        ZF_LOGE("Grabbing the entire cap for device memory");
+        ZF_LOGV("Grabbing the entire cap for device memory");
         err = simple_get_frame_cap(vm->simple, (void *)addr, seL4_PageBits, &frame);
         if (err) {
-            ZF_LOGE("Failed to grab the entire cap");
+            ZF_LOGV("Failed to grab the entire cap for addr 0x%"PRIxPTR, addr);
             return -1;
         }
     }
@@ -1033,6 +1033,7 @@ static vm_frame_t on_demand_iterator(uintptr_t addr, void *cookie)
     /* Attempt allocating device memory */
     err = alloc_vm_device_cap(paddr, vm, &frame_result);
     if (!err) {
+        printf("OnDemandInstall: Created device-backed memory for addr 0x%"PRIxPTR"\n", addr);
         return frame_result;
     }
     /* Attempt allocating ram memory */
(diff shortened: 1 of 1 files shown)


components/VM_Arm/src/main.c
```diff
@@ -1033,13 +1033,15 @@ static vm_frame_t on_demand_iterator(uintptr_t addr, void *cookie)
     /* Attempt allocating device memory */
     err = alloc_vm_device_cap(paddr, vm, &frame_result);
     if (!err) {
+        ZF_LOGE("Created device-backed memory for addr 0x%"PRIxPTR, addr);
```
axel-h, line 1036: Using `ZF_LOGE` is semantically wrong here, this must be `ZF_LOGI`

components/VM_Arm/src/main.c
```diff
@@ -1033,13 +1033,15 @@ static vm_frame_t on_demand_iterator(uintptr_t addr, void *cookie)
...
         return frame_result;
     }
     /* Attempt allocating ram memory */
     err = alloc_vm_ram_cap(paddr, vm, &frame_result);
     if (err) {
         ZF_LOGE("Failed to create on demand memory for addr 0x%"PRIxPTR, addr);
     }
+    ZF_LOGE("Created RAM-backed memory for addr 0x%"PRIxPTR, addr);
```
axel-h, line 1044: Same here, using `ZF_LOGE` is semantically wrong, this must be `ZF_LOGI`

components/VM_Arm/src/main.c
chrisguikema: I'm okay with making this a `ZF_LOGI`. It depends on whether the on-demand info should always be printed or not. Is that information the user always needs to see?

kent-mcleod: The behavior for how and when devices are installed into a vm by the vmm is quite confusing and a main contributor for weird runtime behavior and should be easy to diagnose.

components/VM_Arm/src/main.c
```diff
@@ -992,10 +992,10 @@ static int alloc_vm_device_cap(uintptr_t addr, vm_t *vm, vm_frame_t *frame_resul
     err = vka_utspace_alloc_at(vm->vka, &frame, kobject_get_type(KOBJECT_FRAME, seL4_PageBits), seL4_PageBits, addr,
                                &cookie);
     if (err) {
-        ZF_LOGE("Grabbing the entire cap for device memory");
+        ZF_LOGV("Grabbing the entire cap for device memory");
         err = simple_get_frame_cap(vm->simple, (void *)addr, seL4_PageBits, &frame);
         if (err) {
-            ZF_LOGE("Failed to grab the entire cap");
+            ZF_LOGV("Failed to grab the entire cap for addr 0x%"PRIxPTR, addr);
```
axel-h, line 998: I wonder if the root problem is, that this function should not be used for RAM-backed device install then? Can't this be detected and then this is not tried at all. So the first message can become a `ZF_LOGW()` and this one remains a `ZF_LOGE()`

components/VM_Arm/src/main.c
chrisguikema: I'm going to agree with Kent here. `printf` is used by the VMM in other places to print out important information, and device passthrough is very important. And with the On Demand Device Install, if the device is being backed by RAM, then it may not be working properly, and the user should have at least a hint of what may be causing the problem.

components/VM_Arm/src/main.c
chrisguikema: I think the On Demand order of operations is this:

1. Check for untyped object device capability (if you defined the vm0.dtb {})
2. Check for frame (vm0.mmios)
3. Map the address into RAM

That order makes sense to me. Especially since the On Demand Install is called from a memory fault, and afaik there isn't a way to differentiate between whether that access was for RAM, or for a device.

chrisguikema:
@axel-h @kent-mcleod are we in agreement over using `printf` instead of `ZF_LOG`?

components/VM_Arm/src/main.c
axel-h: Ok, then it's a follow-up task to review the print functions.

chrisguikema:
@kent-mcleod is this good to merge?

kent-mcleod (approved):

Issue #49, opened 2022-10-26
Dont hard-code GIC_IRQ_PHANDLE, read it from device tree instead


axel-h:
Follow-up from https://github.com/seL4/camkes-vm/pull/42#discussion_r1005108779

Remove the hard coded `GIC_IRQ_PHANDLE` define and take the value from the device tree instead. Or add at least a sanity check that the hard-coded values machtes the device tree. 

There have been cases, where the `phande` changes, and that causes strange errors then. 
See also https://github.com/seL4/camkes-vm/issues/27

chrisguikema:
Can we define an `INTERRUPT_CONTROLLER` macro instead? That way you can read that node from the device tree, and get the `phandle`?

chrisguikema:
Is that something that is consistent on a per-platform basis?

Here is the boot core node on a zynqmp device tree:
```
		cpu@0 {
			compatible = "arm,cortex-a53";
			device_type = "cpu";
			enable-method = "psci";
			operating-points-v2 = <0x01>;
			reg = <0x00>;
			cpu-idle-states = <0x02>;
			clocks = <0x03 0x0a>;
			phandle = <0x3d>;
		};
```

On a tx2 device tree, I see `<interrupt-parent>` right on top.

kent-mcleod:
I see.  I guess defining the path is good enough. It could also be used to setup the virtual gic without having to hardcode the addresses for each platform.

chrisguikema:
Created a PR for that here: https://github.com/seL4/seL4_projects_libs/pull/83

axel-h:
Closing this, because https://github.com/seL4/camkes-vm/pull/52 was merged and we use DTB paths now.

PR #52, opened 2022-10-27
vm_arm: use gic path to find phandle

chrisguikema:
This commit defines the gic path for each platform, which is used to parse the device tree to find the gic phandle, instead of relying on the user to provide the correct value.

I could use some help to validate that this functions properly. I did my best to find the proper gic for each platform, but I may have messed things up.

https://github.com/seL4/camkes-vm/issues/49

### DIFF
--- a/components/VM_Arm/plat_include/tx2/plat/vmlinux.h
+++ b/components/VM_Arm/plat_include/tx2/plat/vmlinux.h
@@ -164,7 +164,7 @@ static const int free_plat_interrupts[] = { 220 + GIC_LIC_INTID_BASE };
 #define DTB_ADDR          (LINUX_RAM_BASE + 0x01000000)
 #define INITRD_MAX_SIZE   0x1900000 //25 MB
 #define INITRD_ADDR       (DTB_ADDR - INITRD_MAX_SIZE) //0x80700000
-#define GIC_IRQ_PHANDLE 0x1
+#define GIC_NODE_PATH     "/interrupt-controller@3881000"
 
 static const char *plat_keep_devices[] = {
     "/arm-pmu",
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -802,7 +802,17 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
     }
 
     if (config_set(CONFIG_VM_PCI_SUPPORT)) {
-        err = fdt_generate_vpci_node(vm, pci, gen_fdt, GIC_IRQ_PHANDLE);
+        int gic_offset = fdt_path_offset(fdt_ori, GIC_NODE_PATH);
+        if (gic_offset < 0) {
+            ZF_LOGE("Failed to find gic node from path: %s", GIC_NODE_PATH);
+            return -1;
+        }
+        int gic_phandle = fdt_get_phandle(fdt_ori, gic_offset);
+        if (0 == gic_phandle) {
+            ZF_LOGE("Failed to find phandle in gic node");
+            return -1;
+        }
+        err = fdt_generate_vpci_node(vm, pci, gen_fdt, gic_phandle);
         if (err) {
             ZF_LOGE("Couldn't generate vpci_node (%d)\n", err);
             return -1;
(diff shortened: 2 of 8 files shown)


components/VM_Arm/src/fdt_manipulation.c
```diff
@@ -108,3 +108,23 @@ int fdt_append_chosen_node_with_initrd_info(void *fdt, unsigned long base, size_
 
     return 0;
 }
+
+int fdt_get_gic_phandle(void *fdt, char *gic_path)
+{
+    const fdt32_t *c;
+    int val, len;
```
axel-h, line 115: Seems `val` is not used anywhere?

chrisguikema:
welp theres a `fdt_get_phandle` function that does all this for me...

components/VM_Arm/plat_include/tx2/plat/vmlinux.h
```diff
@@ -164,7 +164,7 @@ static const int free_plat_interrupts[] = { 220 + GIC_LIC_INTID_BASE };
 #define DTB_ADDR          (LINUX_RAM_BASE + 0x01000000)
 #define INITRD_MAX_SIZE   0x1900000 //25 MB
 #define INITRD_ADDR       (DTB_ADDR - INITRD_MAX_SIZE) //0x80700000
-#define GIC_IRQ_PHANDLE 0x1
```
yyshen, line 167: This means that the GIC_NODE_PATH has to match with the name in the DTS, right?

yyshen (approved):

components/VM_Arm/src/main.c
```diff
@@ -802,7 +804,17 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
     }
 
     if (config_set(CONFIG_VM_PCI_SUPPORT)) {
-        err = fdt_generate_vpci_node(vm, pci, gen_fdt, GIC_IRQ_PHANDLE);
+        int gic_offset = fdt_path_offset(fdt_ori, GIC_NODE_PATH);
+        if (gic_offset < 0) {
+            ZF_LOGE("Failed to find gic offset from path: %s", gic_path);
```
axel-h, line 809:
```suggestion
            ZF_LOGE("Failed to find gic node from path: %s", gic_path);
```
Let's says "node" instead of "offset" here.

components/VM_Arm/plat_include/tx2/plat/vmlinux.h
chrisguikema: Correct.

chrisguikema:
Test with: https://github.com/seL4/seL4_projects_libs/pull/83

components/VM_Arm/src/main.c
```diff
@@ -802,7 +804,17 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
     }
 
     if (config_set(CONFIG_VM_PCI_SUPPORT)) {
-        err = fdt_generate_vpci_node(vm, pci, gen_fdt, GIC_IRQ_PHANDLE);
+        int gic_offset = fdt_path_offset(fdt_ori, GIC_NODE_PATH);
```
axel-h, line 807: Should we use `vm.gic_path` here? The file gets initialized in the other code change you did, so this init might have to happen a bit earlier then.

chrisguikema: Order was okay. vm.gic_path is set before this function is called

components/VM_Arm/src/main.c
```diff
@@ -802,7 +804,17 @@ static int generate_fdt(vm_t *vm, void *fdt_ori, void *gen_fdt, int buf_size, si
     }
 
     if (config_set(CONFIG_VM_PCI_SUPPORT)) {
-        err = fdt_generate_vpci_node(vm, pci, gen_fdt, GIC_IRQ_PHANDLE);
+        int gic_offset = fdt_path_offset(fdt_ori, vm->gic_path);
+        if (gic_offset < 0) {
+            ZF_LOGE("Failed to find gic node from path: %s", GIC_NODE_PATH);
```
axel-h, line 809:
```suggestion
            ZF_LOGE("Failed to find gic node from path: %s", vm->gic_path);
```
this also needs an update then

chrisguikema: Done.

axel-h (approved):
Looks good to me now

axel-h:
The builds fail with `‘vm_t’ {aka ‘struct vm’} has no member named ‘gic_path’`, This need a PR in https://github.com/seL4/seL4_projects_libs then also for `libsel4vm/include/sel4vm/guest_vm.h`

chrisguikema:
I'm going to break this up into 2 PRs, so we can get the phandle through, without needing to modify the vm struct.

kent-mcleod (approved):

PR #53, opened 2022-10-27
vm_arm: set fdt pointer and gic name

chrisguikema:
This commit sets the fdt pointer and gic node, which are used by the libraries to find the registers necessary to emulate the gic.

Test with: https://github.com/seL4/seL4_projects_libs/pull/83

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -531,6 +531,8 @@ static int vmm_init(void)
         assert(t_id >= 0);
     }
 
+    camkes_io_fdt(&(_io_ops.io_fdt));
+
     return 0;
 }
 
@@ -882,7 +884,6 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
             close(dtb_fd);
             fdt_ori = (void *)linux_gen_dtb_base_buf;
         } else {
-            camkes_io_fdt(&(_io_ops.io_fdt));
             fdt_ori = (void *)ps_io_fdt_get(&_io_ops.io_fdt);
         }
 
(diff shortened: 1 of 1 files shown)


components/VM_Arm/src/main.c
```diff
@@ -1156,6 +1157,9 @@ int main_continued(void)
     ZF_LOGF_IF(err, "Failed to bind CB to SID");
 #endif /* CONFIG_ARM_SMMU */
 
+    vm.fdt_ori = (void *)ps_io_fdt_get(&_io_ops.io_fdt);
+    vm.gic_path = GIC_NODE_PATH;
```
axel-h, line 1161: Thanks for splitting the PRs up. Please add the usage of `vm->gic_path` in `generate_fdt()` also, so this is the only place where the path's default is set up from the plat constants.

components/VM_Arm/src/main.c
```diff
@@ -1156,6 +1157,9 @@ int main_continued(void)
     ZF_LOGF_IF(err, "Failed to bind CB to SID");
 #endif /* CONFIG_ARM_SMMU */
 
+    vm.fdt_ori = (void *)ps_io_fdt_get(&_io_ops.io_fdt);
```
axel-h, line 1160: Will we place all places where `ps_io_fdt_get()` is calles by `vm->fdt_ori` then also? E.g. in `load_linux()` above?

axel-h:
What is you plan about this one? With https://github.com/seL4/camkes-vm/pull/73 I have another use case where `fdt_ori`in the vm contenxt would be nice, there it's not another global variable flying around.

PR #59, opened 2023-03-04
Upstream/per vm initrd

hlyytine:
Test with: seL4/camkes-vm-examples#35

I see that in seL4/camkes-vm#47 @kent-mcleod was of opinion that changes like this break existing code. I wish to point out in this case (by deprecating ```VmInitRdFile```) we are able to alert everyone at build time, the code will not simply build unless one reads the attached ```docs/deprecated.md``` and makes changes accordingly.

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -865,6 +865,10 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
         if (!initrd || err) {
             return -1;
         }
+        if (initrd_image.size > initrd_max_size) {
+            ZF_LOGE("initrd size %zu over limit (%zu)", initrd_image.size, initrd_max_size);
+            return -1;
+        }
     }
 
     if (!config_set(CONFIG_VM_DTB_FILE)) {
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -66,6 +66,10 @@
 #include <fdtgen.h>
 #include "fdt_manipulation.h"
 
+#if defined(CONFIG_VM_INITRD_FILE)
+#error VmInitRdFile is deprecated, see docs/deprecated.md and fix your code!
+#endif
+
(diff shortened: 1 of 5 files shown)


components/VM_Arm/src/main.c
```diff
@@ -1097,7 +1126,11 @@ int main_continued(void)
     vm_t vm;
     int err;
 
-    parse_camkes_linux_attributes();
+    err = parse_camkes_linux_attributes();
+    if (err) {
+        ZF_LOGF("failed parsing VM attributes");
```
chrisguikema, line 1131: ZF_LOGF stops program flow, right? Change to ZF_LOGE if using the return code.

components/VM_Arm/src/main.c
```diff
@@ -927,15 +933,38 @@ static int load_linux(vm_t *vm, const char *kernel_name, const char *dtb_name, c
...
 {
     linux_ram_base = strtoul(linux_address_config.linux_ram_base, NULL, 0);
     linux_ram_paddr_base = strtoul(linux_address_config.linux_ram_paddr_base, NULL, 0);
     linux_ram_size = strtoul(linux_address_config.linux_ram_size, NULL, 0);
     linux_ram_offset = strtoul(linux_address_config.linux_ram_offset, NULL, 0);
     dtb_addr = strtoul(linux_address_config.dtb_addr, NULL, 0);
+
+    if (strlen(linux_image_config.initrd_name)) {
+        if (!strlen(linux_address_config.initrd_max_size)) {
+            ZF_LOGE("initrd_name is specified, initrd_max_size is not");
```
axel-h, line 946: Do we really need `initrd_max_size`. It feels like an optional parameter to me that is unused in simple cases.

chrisguikema: as far as I can tell, initrd_max_size is completely unused

hlyytine:
Thanks guys for reviewing this, I'm closing this since this is not relevant anymore.

PR #65, opened 2023-03-09
vm_arm: make function static

axel-h:
The function is used internally only, so there is no need to expose it.


axel-h:
Can we merge this to reduce the visibility or is there any hidden dependency here?

kent-mcleod (approved):

PR #63, opened 2023-03-08
vm_arm: use more specific integer types

axel-h:
Basically https://github.com/seL4/camkes-vm/pull/58 from @hlyytine reworked to the recent changes. Best to merge after https://github.com/seL4/camkes-vm/pull/60 is in, then the unused variables `linux_ram_offset` and `initrd_max_size` are gone already

PR #68, opened 2023-03-10
Patch axel 18

axel-h:
Is there a reason why `init_ram.c` exists as a module. To me this seems overly complicated in the current implementation and I do not really understand the benefit for RAM, as it is essential an not optional.


hlyytine:
+1 for removing the init_ram module. But can we make it a weak function, so that VMs could use custom code for overriding how the RAM is mapped?

If you happen to remember our presentation at seL4 summit 2022, we are experimenting with providing virtio backends by Linux application (QEMU, crosvm, you name it) running in another VM. Our first prototype was sharing guest memory to this another VM and in order to do that, we used weak attribute of `init_ram_module()` and provided our own version, which mapped guest physical addresses to dataport frames.

Now I know sharing the whole guest memory is inacceptable in seL4 world and principle of least authority and we working to share only needed pages with SMMU (work in progress) or with software bounce buffers like Linux SWIOTLB (which we have already working). But we would like to maintain this insecure "share all" version as well, to be able to benchmark how much SMMU/SWIOTLB brings overhead.

I guess mapping RAM from dataport frames is not desirable / wanted to be merged into CAmkES-VM as such, but please  provide facilities to be able to override RAM initialization. Weak symbol is the easiest.

axel-h:
I have seen that we were experimenting with this in you fork. What I can do is wrap this in a internal helper function that is weak. What we could als do it on call `vm_ram_register_at()` if `ram_size` is not zero, ie. if there is any exclusive VM RAM. In your case there would be another class of RAM then that can be added via the module hooks, which is just configures in the CMake file then. So there is no need for any `weak` functions.

axel-h:
@hlyytine What do you thing about this implementation?

chrisguikema (commented):
The changes look good to me. 

Is your logic for removing the init_ram module just that its a small function? In my mind, modules should be for optional things. So things like device tree generation, virtio devices, etc. Basically you want to end up with as small of a VMM as possible, right? Since RAM initialization isn't optional (you need it to load in the guest images), I'm okay with removing it as a module.

components/VM_Arm/src/main.c
```diff
@@ -1174,6 +1184,16 @@ int main_continued(void)
         return -1;
     }
 
+    /* Install and initialize guest RAM. There might be no memory here, because
+     * this is the exclusive fixed guest memory. Modules may set up other forms
+     * of guest memory, e.g. memory shared with other VM and memory that is
+     * mapped on demand.
+     */
+    if (ram_size > 0) {
```
chrisguikema, line 1192: if ram_size is 0 or less than zero, we have a problem. I'd change this to:

```
ZF_LOGF_IF(0 > ram_size, "ram_size must be a positive value, as running a guest without RAM is impossible");

err = vm_ram_register_at(&vm, ram_base, ram_size, vm.mem.map_one_to_one);
assert(!err);
```

axel-h: `ram_size` is an `unsigned long`, so this can never be less than zero. With https://github.com/seL4/camkes-vm/pull/63 it will become `size_t`, which also can't ever get negative. We have zero to disable it now. Allowing negative values here has no use case and it seem to just create more trouble.

hlyytine:
> @hlyytine What do you think about this implementation?

Looks good to me now.

components/VM_Arm/src/main.c
chrisguikema: Fair point about the negative, but this would still mean a value of 0 is possible, when it would cause a host of issues.

hlyytine: > The changes look good to me.
> 
> Is your logic for removing the init_ram module just that its a small function? In my mind, modules should be for optional things. So things like device tree generation, virtio devices, etc. Basically you want to end up with as small of a VMM as possible, right? Since RAM initialization isn't optional (you need it to load in the guest images), I'm okay with removing it as a module.

The first part I agree. For example, we have need to customize FDT. I was thinking perhaps we could have `DEFINE_MODULE_EXT` which allows us to specify hooks, for example, during the device tree generation phase. The second that comes to my mind that the fault handler should be extensible with modules. Of course using the MMIO mapping functions is an option, but not sure if we need hooks for `unhandled_mem_fault_callback()` (that's how we have been doing it, but not sure if it's 100% necessary).

I was going to say to I disagree with the second part, but but... @axel-h  how come could we run our "map guest RAM from dataport" before image loading is done? Perhaps hook or weak function is needed after all?

hlyytine: Still some thoughts about modules... I think in more complex systems they will have dependencies so that hooks from module B must be ran before ones from module A. This extended version of `DEFINE_MODULE` macro should handle those as well. Perhaps it needs a little more thinking and solving this with weak function is best for short term solution.

axel-h: > I was going to say to I disagree with the second part, but but... @axel-h how come could we run our "map guest RAM 
> from dataport" before image loading is done? Perhaps hook or weak function is needed after all?

@hlyytine Not sure I get the question. Modules are initialized before images are loaded, so they can provide memory that the modules are put into then. This will not change with this PR. It's just that this RAM init now runs before any module runs. So I don't think a weak hook needs to be provided.

hlyytine: > > I was going to say to I disagree with the second part, but but... @axel-h how come could we run our "map guest RAM
> > from dataport" before image loading is done? Perhaps hook or weak function is needed after all?
> 
> @hlyytine Not sure I get the question. Modules are initialized before images are loaded, so they can provide memory that the modules are put into then. This will not change with this PR. It's just that this RAM init now runs before any module runs. So I don't think a weak hook needs to be provided.

OK, that seems to be case. No need for weak function for that reason and I'm fine with that (provided that you give us a way to generate memory node to DTB). But how about backwards compatibility? That init_ram_module() function has been declared weak since ARM VMM got merged into camkes-vm repo and I wouldn't be surprised if somebody somewhere has overridden it. Shall we summon @kent-mcleod to give his opinion?

hlyytine: > * the memory node is not created at all in the DTB. I consider this more sane than creating a node with a zero size. But in the end I'm collecting input here and would add this in a comment then why we need such a node in any case.

In our use case (map guest RAM from dataport) we need memory node in DTB. We are using autogenerated DTB, so just statically providing it is not enough for us.

axel-h: > In our use case (map guest RAM from dataport) we need memory node in DTB. We are using autogenerated DTB, so just statically providing it is not enough for us.

@hlyytine With the changes in this PR, no "ordinary" memory node is created. Your module would have to create one for the memory it adds. However, this seem impossible in the current code, but I have another PR coming that that allows functions and modules to manipulate the generated DTB according to their "contributions" to the VM. Seems this is exactly what you need then.

axel-h:
@hlyytine I've added a commit that makes the DTB generation more dynamic, so modules can also add data to it. Would this allow you to implement youe use case, if you generate a memory node when `ram_size` is 0?

hlyytine:
> @hlyytine I've added a commit that makes the DTB generation more dynamic, so modules can also add data to it. Would this allow you to implement youe use case, if you generate a memory node when `ram_size` is 0?

Sorry, too many changes for me to understand. Can you look at seL4/camkes-vm#70 and tell me what you think.

 Also I'm not sure if we should do those `if (generate_dtb)` things, perhaps we can also load DTB first (`provide_dtb=1`) and then customize it (`generate_dtb=1`), that way there would be no need for dtb_base_name. Also let DTB loading fail gracefully if `generate_dtb=1`, in that case rely on completely autogenerated DTB.  I guess the latter originated from Dornerworks, so what do you @chrisguikema think, can we imagine simplifying DTB loading code flow like this?

axel-h:
I will move the other commits in this PR to a separate one. Creating the DTB unconditionally is an option, but I  think it's not so nice as it does unnecessarry work any may cause other issues.

hlyytine:
I got a bit hesitant about using `ram_size == 0` notion to disable init_ram module functionality. I think that even we, those familiar with the code, should not rely on our memory. I mean, after successful summer vacation and empty mind, it should be enough to deduce that `ram_size` variable contains the amount of guest RAM. We should not need to remember that `ram_size == 0` means that it does not mean guest does not have any RAM, but rather it is defined somewhere else. Someone might think the guest is emulating XIP ROM and we don't want to steepen seL4 world learning curve, do we?

The second problem is using `ram_size == 0` forces alternative means to introduce yet another configuration variable into CAmkES file. For example in our use case, `ram_size` is the authoritative figure and we just use a CAmkES dataport that is able to provide enough frames to satisfy stage-2 mapping code. It's also kind of self documenting when we set dataport's size to `VMx_RAM_SIZE`.

The third problem, or actually an extension of the second, is that seems that you want to get rid of global variables. Good. I think the reason why we want that is to have pure functions. In other words, to get rid of side effects. Now, enabling or disabling some unknown code based on whether `ram_size` is zero or not does not seem particularly good idea to me. The least we should do is to print a warning message "RAM size is defined as zero, expect some undefined behaviour.

The fourth problem, we have some code (inside seL4 userland libraries) that uses `ram_size` variable. I guess it is not unreasonable to expect somebody else does the same thing somewhere. Now, are we expected to create a new variable for our differently mapped RAM size and patch all the libraries accordingly? Or shall we assign `ram_size` ourselves in our "use CAmkES dataport as RAM" code? If yes to latter, how to communicate this needed fix to all those to might have custom RAM mapping code?

Hence I have a proposition, and this time I will try to explain my thoughts in English instead of PR, but let me address another concern first.

I see you didn't particularly like having multiple module API functions (`init()` and `fdt_generate()` in seL4/camkes-vm#70) and I think you have a point there. I'm fine if we stick with having only `init_module()` entry point, at least for now. But what bothers me is the `void` return type. The only way to react to errors is `assert()` and you cannot propagate the error up the stack to the caller.

1. Leave it as is.
2. Leave it as is and return to this problem later.
3. Introduce an alternative `init()` function with `int` return type and use it if defined, otherwise use the old-style `init_module()` function (in the VMM module loading code). In that case, print warnings also.
4. Add `enum module_state { UNINITIALIZED=0, INITIALIZING=1, FAILED=2, INITIALIZED_IF_RETURNS=3, INITIALIZED=3 }` and add the status field to each `struct vmm_module`. Now before actually calling init_module(), module loading code sets state to `INITIALIZED_IF_RETURNS`. With this we kind of prepare for future when we have proper `int`-returning initializer.
5. Combine 3 and 4.

My personal preference at the moment is 2, followed by 5.

Now back to original problem. The first question is how to allow custom RAM mapping code. I propose that we add a VM attribute `modules` to CAmkES file, defaulting to `all` to retain backwards compatibility. The whitespace separated strings could be prefixed with `-` (remove), otherwise implicit `+` (add) is assumed. Hence `-all` and empty string are equal, no modules loaded (just a theoretical example). More practical example in our use case could be `-init_ram`, our custom module would be included in that "all but init_ram" set. If you need to make sure your VMM has only `init_ram` module, use "-all init_ram". But like I mentioned, if you don't want to be bothered with this, just ignore this, the default `all` makes everything work like earlier.

Optionally, we can add Jinja2 code that checks the `modules` VM attribute and in case some module is explicitly requested in that string but disabled at CMake level, we can issue a warning or halt the build.

Some reasoning about guest RAM: some embedded operating systems don't need DTB and hence we have CAmkES attributes `generate_dtb` and `provide_dtb`. But at hand-written C code level, adding RAM to guest and generating DTB for guest are completely (almost) orthogonal things. The only corner case is that guest does not have any RAM, but even in deeply embedded system with XIP program code in ROM, there is small SRAM for working set. Hence we can drop the almost and say they are completely orthogonal things. So if we are generating DTB, we need to create that memory node also in all cases, regardless whether the RAM is mapped to guest using this code or some custom code. Hence it is reasonable to defend the POV that `ram_base` and `ram_size` global variables could and should exist after some certain of execution flow (installing VMM modules in this case).

_Just a teaser: Therefore, moving DTB generation to module is the most logical conclusion. With Jinja2, we can translate `"generate_dtb" : false` to `-generate_dtb` to be appended to modules. Think about seL4 microkernel. It does the bare minimum only, the rest is done in userspace threads. Why should the VMMs on top of it be of any different design? Shouldn't it just do the bare minimum and delegate the rest to modules (analogy to userspace threads?)_

Lastly, how and when shall we set `ram_base` and `ram_size`. We know that after loading VMM modules they are guaranteed to have sane values and for the remainder of the code, we can take it granted. But where do these values originate? I would imagine that ultimately they are a couple of `#define VMx_RAM_{BASE,SIZE}` just like before but since guest RAM seems to be a required component of a CAmkES VMM, should this configuration be handled through CAmkES interface some day. Then you hook your VM with a CAmkES component that does the same thing as `init_ram` does today, or we at SSRC could hook the VM to a RAM-providing component that is really an extended version of `seL4SharedCapsWithData`. Then my opinions about this:

- Overengineered? Perhaps. But why do we one rather well thought of component system and then additional something more or less hackish one.
- Shall we start implementing this now? No.
- Shall we start coding this in near future? I don't see immediate need.

My opinion concluded: This PR should put on hold, since IMO it really does not improve anything. I suggest that we discuss whether that `modules` VM attribute approach is more suitable.

Action point to me: Start discussion on VMM seL4 discourse.

hlyytine:
I got even simpler idea: seems that we all agree RAM is needed from VM in any case, which means linker must be able to resolve `ram_base` and `ram_size`. Wouldn't it be the most logical place for these global variables to be where the RAM mapping is done, in this case `modules/init_ram.c`? Or in our SSRC case, in our alternative module. That way you would be forced to provide exactly one implementation of RAM mapping code per VM. Provide none, and linker fails due to missing symbol. Provide more than one, and linker fails due to multiple definitions of symbols.

This change would need moving `ram_base` and `ram_size` (and backwards compatibility `linux_ram_ {base,size}`) from `templates/seL4VMParameters.template.h` to a private header `templates/seL4VMRAMConfiguration.h`, which is included by RAM mapping modules only.

We would still need a mechanism to turn off init_ram module and add ours, but other than that, no other changes needed at all.

kent-mcleod:
> I guess mapping RAM from dataport frames is not desirable / wanted to be merged into CAmkES-VM as such, but please provide facilities to be able to override RAM initialization. Weak symbol is the easiest.

I think this was actually why the init_ram module was made.  This example uses a ram dataport to introspect a guest VMs ram device: https://github.com/seL4/camkes-vm-examples/blob/master/apps/Arm/vm_introspect/src/init_dataport_ram.c#L39-L45

kent-mcleod:
> Still one more thought: we really should not compile in code that is not used. Hence perhaps this is best handled at build time at CMake level.

I recall that this simple module system was added to allow general code to be configured in CMake to be compiled in/out based on build time configuration and to also support camkes connectors that use templates that define modules and can thus be initialized at start of component runtime.

I also am not sure about the motivation of this PR is as the module was added for the vm_introspect example to perform its own ram installation.

axel-h:
Thanks. I get the point why this should be a weak function. But I still seems a bit odd to me that this is in a module. It would be a bit more intuitive if the weak function is in `main.c` then with a comment explaining that this is a explicit customization point for modules.

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
     }
 }
 
+/* Default RAM initialization. Modules can overwrite this weak function with a
+ * custom initialization. e.g. to use memory shared with other components or
```
hlyytine, line 624: Perhaps comma instead of dot between "initialization" and "e.g."?

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
...
+ * mapped on demand.
+ */
+WEAK int init_ram_module(vm_t *vm)
```
hlyytine, line 627: Prototype is not the same, `cookie` parameter is missing and return type is different. The latter is more serious; if somebody provided overloaded `init_ram_module` (like in vm-introspect), the return value is of `void` type and hence here the CPU register typically containing return value would be completely random.

While it can be argued that it is hard to see how overloaded versions of `init_ram_module` are able to provide non-NULL cookie, I think if we are really just inlining something, we shouldn't make other changes at the same time.

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
...
+{
+    if (0 == ram_size) {
```
hlyytine, line 629: While I understand the motivation writing comparison in this way, I think it's best to follow the style of existing code. Understanding code is often challenging enough even without mixed style.

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
...
+        ZF_LOGW("VM%d: no RAM defined", vm->vm_id);
```
hlyytine, line 630: Perhaps we should be consistent with existing code and not prefix the message here with `VM%d: `. I once took a look at ZF logging framework and it appears it could be possible to prefix all ZF log messages with some string. While I agree it would be nice to know from which VM the log message is, perhaps we should do it for all messages with that prefix thing?

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
...
+        return 0;
```
hlyytine, line 631: I was under illusion that we agreed that all systems have some RAM, even if very small amount. Now this `0` seems to contradict that?

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
...
+    }
+
+    ZF_LOGI("VM%d: %p - %p (RAM)", vm->vm_id, (void *)ram_base,
```
hlyytine, line 634: 1. Perhaps remove `VM%d: ` prefix as discussed above.
2. Consider using `"0x%"PRIxPTR` instead of `%p`. Then no  need to cast to `(void *)`.
3. How about reporting last valid address instead of first unvalid address?
4. Consistency: error message would complain about registration, but this is just two hexadecimal values?

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
...
+            (void *)(ram_base + ram_size));
+
+    int err = vm_ram_register_at(vm, ram_base, ram_size,
+                                 vm->mem.map_one_to_one);
+    if (err) {
+        ZF_LOGE("RAM registration failed (%d)", err);
```
hlyytine, line 640: For the sake of consistency, I would stick to one of "initialization", "registration" and "installing" and use that consistently.

components/VM_Arm/src/main.c
```diff
@@ -620,6 +620,29 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
...
+        return -1;
```
hlyytine, line 641: How about returning `err` instead of plain -1 where possible?

components/VM_Arm/src/main.c
```diff
@@ -633,7 +656,17 @@ int install_vm_devices(vm_t *vm)
...
     if (config_set(CONFIG_VM_PCI_SUPPORT)) {
         err = vm_install_vpci(vm, io_ports, pci);
-        assert(!err);
+        if (err) {
+            ZF_LOGE("Failed to install VPCI device (%d)", err);
+            return -1;
+        }
+    }
+
+    /* Install and initialize guest RAM. */
```
hlyytine, line 665: See above my comment about choosing one term and sticking to it. I would prefer "initializing".

components/VM_Arm/src/main.c
```diff
@@ -633,7 +656,17 @@ int install_vm_devices(vm_t *vm)
...
+    err = init_ram_module(vm);
```
hlyytine, line 666: As explained above, `cookie` parameter is missing. I think it is enough to pass NULL here.

hlyytine (changes requested):

hlyytine:
> I think this was actually why the init_ram module was made. This example uses a ram dataport to introspect a guest VMs ram device: https://github.com/seL4/camkes-vm-examples/blob/master/apps/Arm/vm_introspect/src/init_dataport_ram.c#L39-L45

Thanks for reminding. Funny enough, that's where I started learning how to share memory between VMs but seems that I forgot it completely. On a more serious note, since there are at least two uses now (vm-introspect and our virtio stuff) one potential idea is to consolidate "RAM as dataport" to a VMM module, but before that, two things need to be solved:  a) per-VM configuration and b) how to tell it which dataport to use.

hlyytine:
> > Still one more thought: we really should not compile in code that is not used. Hence perhaps this is best handled at build time at CMake level.
> 
> I recall that this simple module system was added to allow general code to be configured in CMake to be compiled in/out based on build time configuration and to also support camkes connectors that use templates that define modules and can thus be initialized at start of component runtime.
> 
> I also am not sure about the motivation of this PR is as the module was added for the vm_introspect example to perform its own ram installation.

The trouble seems to be the module here is unnecessary. Weak attribute is the one doing the job. The other alternative is to make `init_ram` a per-VM configurable item, then weak function would not be needed but it would be a API breaking change. Perhaps API is too strong a word, though. If it's not documented, it's not an API, is it?

hlyytine:
@axel-h IMO inlining this would be good, as the weak function kind of negates the purpose of the module system. But how about if you just copied & pasted that function without making any non-whitespace changes. Because it is in the same repo, `git diff` with coloring turned on would show it clearly just as moved code. Then the commit message would explain why the module system does not bring any additional benefits here. If you really insist doing other changes, they could be an another PR.

components/VM_Arm/src/main.c
axel-h: donw, thanks.

components/VM_Arm/src/main.c
axel-h: I've kept this with the comment as sanity check. But I can remove it also completely.

components/VM_Arm/src/main.c
axel-h: I'm not sure if we have global error codes that each caller can make sense on. So that is why I think it is better if each specific function prints the error and then returns a generic error to the caller, as the caller may not make much sende of the specific error code. But in non-debug builds the error code would get lost then, so you have a point there, that we should change this in the whole file to ensure error codes bubble up and there is a change to have them in non-debug builds also at the top level.

components/VM_Arm/src/main.c
axel-h: done.

components/VM_Arm/src/main.c
hlyytine: Another option is to put this check to `templates/seL4VMParameters.template.c`? Putting it in this function and having the option to overload the function would lead someone to think that non-zero RAM size is important only if the RAM is mapped in this way. But we already agree no matter how and where the RAM is mapped from, it needs to be non-zero in size?

And I have to nag about style issue once again. Could you use `vm_config->ram.size == 0` or `!vm_config->ram.size` instead of this, albeit technically superior but not traditional way to write the comparison? If you want the project to use this style, could you start a discussion on seL4 discourse for example and see whether there's support for this in the community? Up until then I think we should not deviate from the existing style.

components/VM_Arm/src/main.c
hlyytine: OK.

components/VM_Arm/src/main.c
```diff
@@ -621,13 +621,36 @@ static void irq_handler(void *data, ps_irq_acknowledge_fn_t acknowledge_fn, void
     }
 }
 
+/* Default RAM initialization. Modules can overwrite this weak function with a
+ * custom initialization, e.g. to use memory shared with other components or
+ * mapped on demand.
+ */
+WEAK int vm_init_ram(vm_t *vm, const vm_config_t *vm_config)
```
hlyytine, line 628: I see you renamed `init_ram_module()` to `vm_init_ram()`. This means at least `vm_introspect` example in https://github.com/seL4/camkes-vm-examples gets broken (it is broken currently for other reasons, but that's another story). Perhaps some code we don't know of.

Personally I'm not against breaking the ABI but others might be.

hlyytine (changes requested):

hlyytine:
To be honest, I think this PR creates more problems than it solves.

I know I repeat myself, but I really believe VMM design should be as minimal as it could be and consolidating the module functionality into `main.c` actually does the opposite.

components/VM_Arm/src/main.c
axel-h: The different name is intentional, it's based on you last comment that changing the signature of an existing function is not a good idea. It this PR gets merged, I will create an issue for `vm_introspect` so this gets adapted.

components/VM_Arm/src/main.c
axel-h: I will create a PR that adds a sanity check in `templates/seL4VMParameters.template.c` and don't print anything here.

axel-h:
> To be honest, I think this PR creates more problems than it solves.

I think this is similar to vPCI, there I also failed for make it purely module because a certain awareness seems a core part of the VMM. As long as we don't have more flexible module API and a proper dependency handling, avoiding hacky modules appears a better was forward for a cleaner VMM.

components/VM_Arm/src/main.c
axel-h: > [`vm_introspect`] is broken currently for other reasons, but that's another story). Perhaps some code we don't know of.

I think https://github.com/seL4/camkes-vm-examples/pull/39 will fix the brokenness you are referring to.

hlyytine: Good spotting but not exactly: at some point CAmkES started using as large as possible frames for dataports. Hence if you had more than 2 MB dataport, it uses ARM large pages. On the other hand, currently VMM always uses 4 kB pages in `vm_ram_touch()`. To be able to use 2 MB pages from dataport as guest RAM, we use this kind of hack in tiiuae/sel4_projects_libs@fa99679290c03deba72b18ac39b491599eb63d31 .

hlyytine: Forgot to add that you need to use large pages here as well: https://github.com/seL4/camkes-vm-examples/blob/master/apps/Arm/vm_introspect/src/init_dataport_ram.c

axel-h: @hlyytine: Could you create an issue in https://github.com/seL4/camkes-vm-examples to track this?

PR #69, opened 2023-03-10
vm_arm: add priority in vm config struct 

axel-h:
This builds in top of https://github.com/seL4/camkes-vm/pull/62 and adds the priorities to the vm config struct  from the template. Is also make the `VM_INIT_DEF` use 100 as default base priority, so there is no need to explicitly specify anything. It also support the `_priority`. However, I wonder if we should clean this up an rename it to ìrqserver_prio`in the attribute also in a follow-up task


templates/seL4VMParameters.template.c
```diff
@@ -8,17 +8,26 @@
 #include <camkes.h>
 #include <arm_vm/gen_config.h>
 
-/*- set vm_address_config = configuration[me.name].get('vm_address_config') -*/
-/*- set vm_image_config = configuration[me.name].get('vm_image_config') -*/
-/*- set linux_address_config = configuration[me.name].get('linux_address_config') -*/
-/*- set linux_image_config = configuration[me.name].get('linux_image_config') -*/
+/*- set config = configuration[me.name] -*/
```
hlyytine, line 11: I'm wondering whether these deserve a commit of their own (even in this very same PR); good stuff, but not exactly about priorities.

hlyytine:
Also, shouldn't you get rid of hard coded value at https://github.com/seL4/camkes-vm/blob/ce598e47ef9df054e2d160bdd3fd51af07672686/components/VM_Arm/vm_common.camkes#L33 (and the other one on line 35 also)?

templates/seL4VMParameters.template.c
axel-h: moved to a separate commit

components/VM_Arm/src/main.c
```diff
@@ -1209,15 +1207,20 @@ static int main_continued(void)
...
 {
-    /* if the base_prio attribute is set, use it */
-    if (&base_prio != NULL) {
-        VM_PRIO = base_prio;
+    if (vm_config.priority.irqserver <= vm_config.priority.vmm) {
+        ZF_LOGF("IRQ server priority (%u) must be higher than VMM priority (%u)",
+                vm_config.priority.irqserver, vm_config.priority.vmm);
     }
+
+    if (vm_config.priority.vmm <= vm_config.priority.vcpu) {
```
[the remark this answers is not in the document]
axel-h: I've bee wondering about where the right place for the check should be. Having it in the template seems to make more sense for now.

axel-h:
> Also, shouldn't you get rid of hard coded value at camkes-vm/components/VM_Arm/vm_common.camkes

Yes, seems this can also be removed then.

hlyytine (approved):
Looks good to me.

components/VM_Arm/vm_common.camkes
```diff
@@ -30,9 +30,6 @@ assembly {
         vm.cnode_size_bits = 23;
         vm.simple_untyped24_pool = 12;
 
-        vm.base_prio = 100;
-
-        vm._priority = 101;
```
kent-mcleod, line 35: `_priority` has a special meaning and is the default priority for the thread that calls `run`

axel-h: Thanks, worth to add this in a comment. But shouldn't this be 100 then?

kent-mcleod: If the VMM threads aren't a higher priority than the vm thread, the vm won't be preempted when there's work to be done in the VMM which affects things like IRQ latency.

axel-h: But the "VM Thread" is the vCPU thread, right? This is supposed to run at "basePrio - 1".

kent-mcleod: I guess it could still be up to someone to choose to configure the vmm and vcpu threads with the same priority if they wanted to delay things like IRQ injection until the vcpu thread was preempted or yielded, but in general, it's probably most straight forward if the irq server threads have equal or higher prio to vmm thread, which has higher prio to vcpu threads.

PR #70, opened 2023-03-11
Upstream/module hooks

hlyytine:
- This makes it possible for a VMM module to customize generated/provided DTB.
- Existing APIs are not changed.
- Existing semantics are not changed, existing code should compile/work as earlier.
- The same pattern can be used for other hooks in the future.

### DIFF
--- a/components/VM_Arm/include/vmlinux.h
+++ b/components/VM_Arm/include/vmlinux.h
@@ -30,11 +30,14 @@ struct irq_token {
 };
 typedef struct irq_token *irq_token_t;
 
+typedef int (*vmm_module_fdt_generate_fn_t)(void *fdt, vm_t *vm, void *cookie);
+
 typedef struct vmm_module {
     const char *name;
     void *cookie;
     // Function called for setup.
     void (*init_module)(vm_t *vm, void *cookie);
+    vmm_module_fdt_generate_fn_t *fdt_generate;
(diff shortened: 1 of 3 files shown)


hlyytine:
The proposal from @chrisguikema that VMM should be modular and minimizable resonates well within our team. I found tiiuae/camkes-vm#68 from @axel-h particularly hard to follow (read: not sure if I understood everything). Can you look at this and tell me what you think.

Note that if we go with this approach, we shall fix that TODO before merging, i.e., move stuff from fdt_manipulation.c to external library (or more specifically, provided externally usable header) for modules to use.

hlyytine:
And yes, this is not an exact alternative to seL4/camkes-vm#68: it does not have mechanism how to disable init_ram module and let some other module map the guest RAM. I have some ideas, I just need to write them down.

axel-h:
I think this makes the module interface even more complex. My preference is, that there is one function that gets a context pointer with all VMM/VM things. Then this is a clean OO approach without global variables. I will make a proposal for that.
The module dependency issue might something separate to think about still.

hlyytine:
Yeah, after all FDT generation can itself be a module; if and when we add mechanism for modules to require some other modules to be loaded, optional FDT generation could be their own `DEFINE_MODULE`.

hlyytine:
Actually my last comment is wrong. The DTB generation code is only after loading modules.

PR #71, opened 2023-03-13
vm_arm: add helper function add_uts(), fix corner case bug

axel-h:
This also handles the special case where RAM size is zero correctly, for base=0 and size=0 it resulted in 'true' for any address before.

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -395,6 +395,61 @@ static int camkes_vm_utspace_alloc_at(void *data, const cspacepath_t *dest, seL4
 
 }
 
+static bool add_uts(const vm_config_t *vm_config, vka_t *vka, seL4_CPtr cap,
+                    uintptr_t paddr, size_t size_bits, bool is_device)
+{
+    cspacepath_t path;
+    vka_cspace_make_path(vka, cap, &path);
+
+    /*
+     * The general usage concept for the different UT pools is:
+     *
+     * ALLOCMAN_UT_KERNEL:
+     *   Physical RAM mapped in the kernel windows. This can be used to create
+     *   arbitrary kernel objects.
+     *
+     * ALLOCMAN_UT_DEV_MEM:
+     *   UTs from a device region, which is also usable RAM. This exists because
+     *   on 32-bit platforms the kernel window could be too small to map all
+     *   physical RAM and thus is able to directly access it. Kernel objects can
+     *   be created from these UTs, if the 'canBeDev' parameter in 'alloc' is
+     *   set to true. However, such objects may have restrictions in what it can
+     *   be used for, because the kernel cannot access the content directly.
+     *
+     * ALLOCMAN_UT_DEV:
+     *   Device regions. Such UTs will never be used for an allocation, unless
+     *   explicitly requested by physical address.
+     *
+     * ToDo: The practical distinction between ALLOCMAN_UT_DEV and
+     *       ALLOCMAN_UT_DEV_MEM is, whether the UT can be requested from the
+     *       allocator without providing it's backing physical address. Since
+     *       UTs from the guest RAM region are supposed to be requested using
+     *       the physical address anyway, there seem not reason why they can't
+     *       be in the ALLOCMAN_UT_DEV pool, too.
+     *       The only know use case where this matters is on the NVidia TK1
+     *       platform. It has a SMMU, so in order to use RAM in a VM, there is
+     *       no need for VMs to have their RAM addresses match the physical RAM
+     *       addresses (VM config option "map_one_to_one"). However, it is a
+     *       32-bit architecture and the kernel window is too small to cover all
+     *       physical RAM. Thus, the additional RAM is made available via device
+     *       UTs. The allocator needs to be told about this, which is what
+     *       "ram_paddr_base" is used for and and why it's different from
+     *       "ram_base".
+     */
+
+    bool is_guest_ram = (paddr >= vm_config->ram.phys_base) &&
+                        ((paddr - vm_config->ram.phys_base) < vm_config->ram.size);
+
+    int ut_type = !is_device ? ALLOCMAN_UT_KERNEL
+                  : is_guest_ram ? ALLOCMAN_UT_DEV_MEM
+                  : ALLOCMAN_UT_DEV;
+
+    allocman_t *allocman = vka->data;
+
+    return allocman_utspace_add_uts(allocman, 1, &path, &size_bits, &paddr,
+                                    ut_type);
+}
+
 static int vmm_init(const vm_config_t *vm_config)
 {
     vka_object_t fault_ep_obj;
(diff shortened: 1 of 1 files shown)


components/VM_Arm/src/main.c
```diff
@@ -445,8 +454,7 @@ static int vmm_init(void)
         cspacepath_t path;
         vka_cspace_make_path(vka, cap, &path);
         int utType = device ? ALLOCMAN_UT_DEV : ALLOCMAN_UT_KERNEL;
-        if (utType == ALLOCMAN_UT_DEV &&
-            paddr >= ram_paddr_base && paddr <= (ram_paddr_base + (ram_size - 1))) {
+        if (utType == ALLOCMAN_UT_DEV && is_paddr_in_vm_ram(paddr)) {
```
hlyytine, line 457: How about just `paddr >= ram_paddr_base && paddr < ram_paddr_base + ram_size`?

hlyytine:
Once again, what do we mean by "VM RAM"? VM = VMM + guest, isn't it? So perhaps we should call it the guest RAM...

... but the problem is that these loops check whether `paddr` falls within the range of physical addresses reserved for stage-2 mapping the _guest physical address_ space from. So the function name `is_paddr_in_vm_ram()` makes me think of the old adage "comparing apples to oranges". So if you really want to use a helper function, how about a function like `static bool reserved_for_guest(uintptr_t paddr)`? That's want we really wanted to know, right?

Previously I made a comment that just fix the condition inline, but I might want to take that comment back. You see, there is another, albeit unrelated issue with `map_one_to_one`, or disregarding it here. Even though if it were `false`, and VMM was given free hands to use whatever untyped memory to back up the guest vspace stage-2 mapping, you still need to provide `ram_paddr_base`, which is clearly wrong.

I have not pondered how to fix that properly (read: I haven't studied all the details of that allocator), but I was thinking if you added that `reserved_for_guest()` function, we could add support for `map_one_to_one` in one central location later on.

components/VM_Arm/src/main.c
axel-h: That is also an option. Having a helper function avoids some redundancy. This comes handy in a few other PR also.

kent-mcleod (dismissed):

kent-mcleod:
> Previously I made a comment that just fix the condition inline, but I might want to take that comment back. You see, there is another, albeit unrelated issue with `map_one_to_one`, or disregarding it here. Even though if it were `false`, and VMM was given free hands to use whatever untyped memory to back up the guest vspace stage-2 mapping, you still need to provide `ram_paddr_base`, which is clearly wrong.

It might be possible to simply remove these comparisons all-together.  The distinction between `ALLOCMAN_UT_DEV` and `ALLOCMAN_UT_DEV_MEM` is whether the UT can be requested from the allocator without providing it's backing physical address. Assuming that we're using the ram_base + size region to mark any untypeds we see as `ALLOCMAN_UT_DEV_MEM`, only to request them later using the physical address anyway, it would be fine to just leave them in the `ALLOCMAN_UT_DEV` pool, which may get checked after the `ALLOCMAN_UT_DEV_MEM` pool anyway?

axel-h:
> It might be possible to simply remove these comparisons all-together. 

I don't know the VMM details good enough yet to have an answer here. It seems this would change the behavior slightly, so it might potentially affect modules then? It might be a change to do after this PR then.

kent-mcleod:
> I don't know the VMM details good enough yet to have an answer here. It seems this would change the behavior slightly, so it might potentially affect modules then? It might be a change to do after this PR then.

Then it might be easier to give the helper a very specific name for now: `is_ut_cap_in_static_ram_region` or something?

axel-h:
Thanks for the feedback. This is now starting to move slightly away from my initial motivation to have the helper function. But it seems using a more specific helper function with some comments could really help making the code better to understand.

components/VM_Arm/src/main.c
```diff
@@ -395,6 +395,46 @@ static int camkes_vm_utspace_alloc_at(void *data, const cspacepath_t *dest, seL4
 
 }
 
+static bool add_uts(vka_t *vka, seL4_CPtr cap, uintptr_t paddr, size_t size,
+                    bool is_unrestricted)
```
kent-mcleod, line 399: is_device is probably a better flag to use as it more closely aligns with both the `simple_` interface and the `seL4_Bootinfo_t` untyped definitions?

kent-mcleod:
So these comparisons are currently depended on by the tk1 platform.  It has an SMMU and doesn't need to directly map the stage-2 translations, however as a 32bit architecture the kernel window is too small to hold all of RAM and so extra ram is available as device untyped, however the allocator needs to be told about this which is what `linux_ram_paddr_base` is and why it's different from `linux_ram_base`.  Hopefully this answers some of @hlyytine's questions?

components/VM_Arm/src/main.c
axel-h: Good point, I've changed it to use `is_device`.

hlyytine:
> Hopefully this answers some of @hlyytine's questions?

It does, thanks.

components/VM_Arm/src/main.c
```diff
@@ -395,6 +395,67 @@ static int camkes_vm_utspace_alloc_at(void *data, const cspacepath_t *dest, seL4
...
+     *   explicitly requested by physical address.
+     *
+     * ToDo: The practical distinction between ALLOCMAN_UT_DEV and
+     *       ALLOCMAN_UT_DEV_MEM is, whether the UT can be requested from the
+     *       allocator without providing it's backing physical address. Since
+     *       UTs from the guest RAM region are supposed to be requested using
+     *       the physical address anyway, there seem not reason why they can't
+     *       be in the ALLOCMAN_UT_DEV pool, too.
+     *       The only know use case where this matters is on the NVidia TK1
+     *       platform. It has a SMMU, so in ordere to use DAM in a VM, there is
```
hlyytine, line 432: "ordere", "DAM"

axel-h: Thanks, fixed.

components/VM_Arm/src/main.c
```diff
@@ -395,6 +395,67 @@ static int camkes_vm_utspace_alloc_at(void *data, const cspacepath_t *dest, seL4
...
+     *       addresses (VM config option "map_one_to_one"). However, it is a
+     *       32-bit architecture and the kernel window is too small to cover all
+     *       physical RAM. Thus, the additional RAM is made available via device
+     *       UTs. The allocator needs to be told about this, which is what
+     *       "ram_paddr_base" is used for and and why it's different from
+     *       "ram_base".
+     */
+
+    bool is_guest_ram = false;
+    if (ram_size > 0) {
```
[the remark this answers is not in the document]
axel-h: We can also use this expression, as it seems overflow safe.
I like the "?:" expression, because you have only one assignment to a variable then. But I expect the compiler to generate the same code from each expression.

kent-mcleod (approved):

PR #73, opened 2023-03-13
vm_arm: make DTB generation more flexible

axel-h:
Allow generating DTB on the fly when the VM is created/manipulated. This also allows modules to manipulate it.
This partly overlaps with what https://github.com/seL4/camkes-vm/pull/70 is doing.

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -865,22 +865,14 @@ static int load_generated_dtb(vm_t *vm, uintptr_t paddr, void *addr, size_t size
     return 0;
 }
 
-static int load_vm(vm_t *vm, const vm_config_t *vm_config)
+static int load_vm_images(vm_t *vm, const vm_config_t *vm_config)
 {
     seL4_Word entry;
     seL4_Word dtb;
     int err;
 
-    /* Install devices */
-    err = install_vm_devices(vm);
-    if (err) {
-        printf("Error: Failed to install VM devices\n");
-        return -1;
-    }
-
-    printf("Loading Kernel: \'%s\'\n", vm_config->files.kernel);
-
     /* Load kernel */
+    printf("Loading Kernel: \'%s\'\n", vm_config->files.kernel);
     guest_kernel_image_t kernel_image_info;
     err = vm_load_guest_kernel(vm, vm_config->files.kernel, vm_config->ram.base,
                                0, &kernel_image_info);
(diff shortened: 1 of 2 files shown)


axel-h:
Seems that with this change the virtual PCI device can be moved out of the VMM code into a module.

hlyytine (approved):
Makes sense and looks good to me. Would it be worth adding
```
#define DTB_BUFFER_SIZE 0x50000
extern char gen_dtb_buf[DTB_BUFFER_SIZE];
```
to `components/VM_Arm/include/vmlinux.h`? That way modules could use that `gen_dtb_buf` without having to declare it themselves.

axel-h:
> Makes sense and looks good to me. Would it be worth adding
> 
> ```
> #define DTB_BUFFER_SIZE 0x50000
> extern char gen_dtb_buf[DTB_BUFFER_SIZE];
> ```

Yes, thanks for the note.

axel-h:
Turns out, we need both `gen_dtb_buf` and `fdt_ori` to be public. See https://github.com/seL4/camkes-vm/pull/77

components/VM_Arm/src/main.c
```diff
@@ -754,66 +787,66 @@ static int generate_fdt(vm_t *vm, const vm_config_t *vm_config,
...
-    if (err) {
-        ZF_LOGE("Couldn't generate chosen_node (%d)\n", err);
+    return 0;
+}
+
+static int vm_dtb_finalize(vm_t *vm, const vm_config_t *vm_config)
+{
+    assert(vm_config->generate_dtb);
+
+#ifdef CONFIG_VM_PCI_SUPPORT
```
kent-mcleod, line 828: Why has `if (config_set(CONFIG_VM_PCI_SUPPORT)) {` been removed?

kent-mcleod (approved):

components/VM_Arm/src/main.c
axel-h: I can use this also. It's a leftover from figuring out the dependencies.

lsf37:
This PR seems to have broken the `camkes-vm-examples` and `sel4-webserver` applications (sel4-webserver probably jus because it uses a VM). Breaking run is here:

https://github.com/seL4/camkes-vm-examples/actions/runs/4661009942/jobs/8250643632#step:4:407

```
 Booting all finished, dropped to user space
  run@main.c:1296 Invalid 'num_vcpus' attribute setting: Exceeds maximum number of supported nodes. Capping value to CONFIG_MAX_NUM_NODES (1)
  clean_up@fdtgen.c:374 Non-existing node /pmu@7000a000 specified to be kept
  clean_up@fdtgen.c:374 Non-existing node /i2c@7000d400 specified to be kept
  clean_up@fdtgen.c:374 Non-existing node /i2c@7000d600 specified to be kept
  clean_up@fdtgen.c:374 Non-existing node /i2c@7000d800 specified to be kept
  clean_up@fdtgen.c:374 Non-existing node /i2c@7000da00 specified to be kept
  FAULT HANDLER: data fault from vm0.vm0_0_control (ID 0x2) on address 0x90, pc = 0x4646a8, fsr = 0x93800006
```

lsf37:
I'll have a look at adding HW actions for this repo, we do have enough infrastructure available for that now.

PR #75, opened 2023-03-20
vm_arm: check return codes properly

axel-h:
- `simple_get_untyped_count()` can return an error code.
- `camkes_dtb_untyped_count()` may also return an error, given the current signature does not specify it only returns unsigned integers.

### DIFF
--- a/components/Init/src/main.c
+++ b/components/Init/src/main.c
@@ -198,7 +198,9 @@ void pre_init(void)
     }
 
     /* add untyped mmios */
-    for (int i = 0; i < simple_get_untyped_count(&camkes_simple); i++) {
+    int cnt = simple_get_untyped_count(&camkes_simple);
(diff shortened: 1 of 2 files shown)


components/Init/src/main.c
```diff
@@ -198,7 +198,9 @@ void pre_init(void)
     }
 
     /* add untyped mmios */
-    for (int i = 0; i < simple_get_untyped_count(&camkes_simple); i++) {
+    int cnt = simple_get_untyped_count(&camkes_simple);
+    assert(cnt >= 0); /* This shall never fail. */
```
kent-mcleod, line 202: It might be better to use ZF_LOG_IF(cnt < 0, "some error message") to explicitly abort on failure even if asserts are disabled?

axel-h: Changed.

kent-mcleod (approved):

PR #77, opened 2023-03-31
vm_arm: move vpci code to dedicated module 

axel-h:
This PR is on top of https://github.com/seL4/camkes-vm/pull/73 and uses the DTB generation refactoring to move the vpci support imto a dedicate module.

### DIFF
--- a/arm_vm_helpers.cmake
+++ b/arm_vm_helpers.cmake
@@ -45,10 +45,17 @@ function(DeclareCAmkESARMVM init_component)
         vm_src
         ${ARM_VM_PROJECT_DIR}/components/VM_Arm/src/main.c
         ${ARM_VM_PROJECT_DIR}/components/VM_Arm/src/fdt_manipulation.c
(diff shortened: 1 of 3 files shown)


hlyytine:
I'm working on generalizing that cross-VM connector. My intention is to support mapping as RAM and providing MMIO-style interface in addition to that PCI interface. So perhaps `crossvm.c` should not depend on PCI support that hard.

axel-h:
> I'm working on generalizing that cross-VM connector. My intention is to support mapping as RAM and providing MMIO-style interface in addition to that PCI interface. So perhaps `crossvm.c` should not depend on PCI support that hard.

It would take another PR for this file to relax the dependency then. At the moment that seem a big practical dependency, that is why I made this change.

axel-h:
After some more refactoring, making a VPCI module seems not really useful, as this would be just a wrapper around the vPCI helper coede in `libsel4vmmplatsupport` then. The refactoring seem an improvement, though.

PR #81, opened 2023-04-11
Debug/Fix seL4/camkes-vm#73

axel-h:
Debug/Fix seL4/camkes-vm#73

### DIFF
--- a/components/VM_Arm/src/main.c
+++ b/components/VM_Arm/src/main.c
@@ -863,12 +863,6 @@ static int vm_dtb_init(vm_t *vm, const vm_config_t *vm_config)
 
     /* Now the DTB is in gen_dtb_buf and all manipulation must happen there. */
 
-    err = fdt_generate_plat_vcpu_node(vm, gen_dtb_buf);
-    if (err) {
-        ZF_LOGE("Couldn't generate plat_vcpu_node, error %d", err);
-        return -1;
-    }
-
     /* generate a memory node */
     err = fdt_generate_memory_node(gen_dtb_buf, vm_config->ram.base,
                                    vm_config->ram.size);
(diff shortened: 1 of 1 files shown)


kent-mcleod:
I can check if this runs on the tx1?

.github/workflows/test.yml
```diff
@@ -18,6 +18,47 @@ jobs:
...
         march: ${{ matrix.march }}
+    - name: Upload images
+      uses: actions/upload-artifact@v2
+      with:
+        name: images-${{ matrix.march }}
+        path: '*-images.tar.gz'
+
+  hw-test:
+    name: HW Test
+    runs-on: ubuntu-latest
```
lsf37, line 33: This one isn't going to work like this, unfortunately. We need to change the trigger to `pull_request_target` to access secrets and also adding labelling etc to the trigger for it to be called then. It then has to be merged first so that `pull_request_target` can become active.

Also, the camkes-vm tests didn't actually fail, it was the camkes-vm-examples tests that failed. I can have a look later at adding label triggers for both of them for PRs on the camkes-vm  repo (will take a bit, I'm in meetings the rest of the day).

kent-mcleod:
I just manually checked and this change resolves the crashes on tx1 and tk1.

kent-mcleod:
Can you explain what the issue was that was introduced by the other PR?

axel-h:
> Can you explain what the issue was that was introduced by the other PR?

I don't really know yet. Moving `fdt_generate_plat_vcpu_node()` was expected to avoid an error seen in one case, but that is also fixed the crashes is quite strange.

axel-h:
> Can you explain what the issue was that was introduced by the other PR?

Yes, not I figured it out.. When creating the vCPU device tree nodes, the code in https://github.com/seL4/seL4_projects_libs/blob/master/libsel4vmmplatsupport/src/arch/arm/guest_vcpu_util.c#L100 blindly dereferences the vcpu pointer. But it's not set yet set as `vm_dtb_init()`, as this is called quite early in the VMM/VM setup process. It does not crash in some cases because `vm->num_vcpus` happens to be 0 and thus the loop in `fdt_generate_plat_vcpu_node()` has no iteration.
Calling `fdt_generate_plat_vcpu_node()` as part of `vm_dtb_finalize()` seems the best option, as the VMM/VM setup code and also modules have the option to manipulate the vCPU details then.

But I still see failures, so there seems another issue.

lsf37:
(Now needs rebase since #82 is merged)

lsf37:
Ok, it's still testing the master branch, because I forgot to provide the PR sha (see #84).

axel-h:
I still need this to fix the issue introduced in https://github.com/seL4/camkes-vm/pull/73

axel-h:
OK, seem github closed this when the base branch was deleted.

lsf37:
Hm, yes, it looks like I can't reopen it either, sorry. If you rebase it on master, maybe it can be reopened then?

lsf37:
Right, just saw #87, all good.

PR #82, opened 2023-04-12
github: PR hardware tests for camkes-vm-examples

lsf37:
Add a label trigger for hardware tests of camkes-vm-examples for PRs to the camkes-vm repository.

(Very similar to @axel-h's proposal in #81)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-# Copyright 2021, Proofcraft Pty Ltd
+# Copyright 2023, Proofcraft Pty Ltd
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
@@ -7,11 +7,29 @@
 name: Camkes VM Examples
 
 on:
-  pull_request:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled]
+
+# downgrade permissions to read-only as you would have in a standard PR action
+permissions:
+  contents: read
 
 jobs:
-  test:
+  code:
+    name: Freeze Code
+    runs-on: ubuntu-latest
+    outputs:
+      xml: ${{ steps.repo.outputs.xml }}
+    steps:
+    - id: repo
+      uses: seL4/ci-actions/repo-checkout@master
+      with:
+        manifest_repo: camkes-vm-examples-manifest
+        manifest: master.xml
+
+  build:
     name: Build
+    needs: [code]
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


lsf37:
This seems to be working fine and shows the failing hardware tests, so I'll remove the testing commit again such that the PR is mergable.

axel-h (approved):
Looks ok to me, thanks!

.github/workflows/test.yml
```diff
@@ -20,4 +38,45 @@ jobs:
...
+    - name: Upload images
+      uses: actions/upload-artifact@v3
+      with:
+        name: images-${{ matrix.march }}
+        path: '*-images.tar.gz'
+
+  hw-run:
+    name: Hardware
+    runs-on: ubuntu-latest
+    needs: [build]
```
axel-h, line 52: Is there any specific reason why you use the brackets here, but not in line 32 ("` needs: code`")? Seems It's copy/pasted over several repos, so the the original reason might have been that there was more in this array in another repo. We could you the array syntax everywhere for consistency.

axel-h (commented):
It seem this does no pick up the PR change, but uses just the base branch for the build and test now.
For https://github.com/seL4/camkes-vm/pull/81 I see in the build run https://github.com/seL4/camkes-vm/actions/runs/4691000562/jobs/8314886188
```
  Starting repo checkout on branch master with manifest master.xml:
  
  repo has been initialized in /github/workspace
  Using supplied manifest XML
  repo: reusing existing repo client checkout in /github/workspace
  
  repo has been initialized in /github/workspace
  repo sync has finished successfully.
  Manifest summary:
  -----------------
  camkes-vm-examples-manifest: 6d870c12 Updating default.xml                           (origin/master)
                         seL4: fb808f05 boot: sanity checks for provide_untyped_cap .. 
                  camkes-tool: d7d37df9 cmake: improve maintainability                 
             camkes-vm-images: 69af5316 github: trigger main test on push to master .. 
                        capdl: 3fa3a1b5 trivial: fix style for changed file            
            global-components: 908195f6 SerialServer: add ^C and ^Z to switching logic 
                       libzmq: d062edd8 Finalise changelog for 4.2.5                   
                         lwip: 159e31b6 Prepare 2.1.2 release                          (tags/STABLE-2_1_2_RELEASE^0)
                     musllibc: 3d6b939e aarch64_sel4: Use more supported asm directive 
                      picotcp: 13c00a06 minor README.md text improvements              
                projects_libs: 524777b7 trivial,libvswitch: Style include/vswitch.h    
                    seL4_libs: 2ca52542 libsel4bench: add support for ARM Cortex-A55   
           seL4_projects_libs: e919ff93 libsel4vmmplatsupport: remove unrechable ret.. 
                  sel4runtime: d935dd05 aarch32: Rename __aeabi_read_tp file (#15)     
                    util_libs: e8e30909 libcpio: fix conversion of -1 to ulong         
                    camkes-vm: 99e9f708 github: PR hardware tests for camkes-vm-exam.. 
           camkes-vm-examples: 9f0a5344 Add documentation of package structure         
              camkes-vm-linux: 46fb8b6e vm-linux-helpers: symbolic links in overlays   
                        polly: ef7e79c2 README.md                                      
                   seL4_tools: 2c213a9e elfloader: don't shadow variable       
```
There `camkes-vm` should be commit 86e42bf or something. Or is the problem here already, that I'm not planning to merge to `master`, but to 99e9f708 (which add the hw test feature) as base?

.github/workflows/test.yml
lsf37: Both work, but yes always using a list would be more consistent.

lsf37:
Let me merge this, then we can check if the problem still persists. I would have expected the PR checkout code to work, but I don't think we have tested that scenario yet.

axel-h:
Is there a way to figure out why for https://github.com/seL4/camkes-vm/actions/runs/4716178344 the hw tests were not running? I could not find any logs that show some insights. The condition check for hw runs seems the same as in the sel4 kernel, so they should work. The label on the PR was set already. All I have changed was removing the debug log commits form the branch  and set a "bug" label. So my guess is, that setting the "bug" label somehow caused this behavior. 

Anyway, removing and setting the "hw-test" label fixes this. However, the build then runs again instead of just re-running the hw test.

lsf37:
When the event is `labeled`, which it would be for setting `bug`, then the build is restarted (because it is unconditional), but the hardware run is not restarted, because you're setting an unrelated label. I agree that it looks confusing, but I think this is actually fine, since nothing in the code has changed on that event, so logs for the previous hardware run should still exist and be valid and there is no reason to rerun them.

The main problem is that the Checks summary on the PR only shows the most recent run -- in your case that was where the hardware jobs were skipped. If you go to the "Checks" tab instead (on https://github.com/seL4/camkes-vm/pull/87/checks), you can see that for the latest commit there are 3 groups of Camkes VM Examples jobs, 2 where the hardware tests ran (the first and the last), and one where it didn't run (the second, which was triggered by the bug label).

PR #87, opened 2023-04-16
Fix crashes introduced in seL4/camkes-vm#73

axel-h:
This fixes the crashes introduced with the changes from commit fb3023b7 from seL4/camkes-vm#73


axel-h:
Looks like e9b5bd5 could be the proper fix things in this repo. Furthermore, the libs could get some sanity checks, so they report an error instead of crashing on a null pointer, I'll make a PR there then also.

lsf37 (approved):
That looks reasonable to me. @kent-mcleod any concerns? Would be good to unbreak the build again.

kent-mcleod (approved):

PR #90, opened 2023-04-20
vm_arm: define loop variable in loop header

axel-h:
Code cleanup to make clear there is no additional dependency on the loop variable.


axel-h:
Can we merge this?

kent-mcleod (approved):

PR #91, opened 2023-04-20
vm_arm:  cleanup and improve documentation

axel-h:


axel-h:
Can we merge this? 

I wonder, if the `DEPENDS` parameters is something to be removed one day, as I can't really come up with a use case why to put something here - except that this allows solving a problem that is somewhere else actually. At least that seems to be the case for all the usages that we have in the current code base. Maybe these were necessary in older version of CMake or they are just over cautious. Anyway, the verbose explanation should help for now to state what we guarantee here.

camkes_vm_helpers.cmake
```diff
@@ -257,20 +261,22 @@ endfunction(DeclareCAmkESVMRootServer)
 # Parameters:
 #
 # <filename_pref>
-#   The name the caller wishes to use to reference the file in the CPIO archive.
-#   This corresponds with the name set in the 'kernel_image' camkes variable for
-#   a given instance vm.
+#   The name to use for the file in the file server
```
lsf37, line 264: Is the part about kernel_image etc no longer true? If not I'd leave that in to make the connection clearer.

lsf37 (commented):
(otherwise this looks good from my side)

camkes_vm_helpers.cmake
axel-h: I found it more confusing that helpful to mention this here. It seems that `kernel_image` is only used for the x86 VMM, the ARM VMM uses either `linux_image_config.linux_name` (old, deprecated) or `vm_image_config.kernel_name`(new) now. The VMMs are diverged a bit, unifying things might be worth a thought...

axel-h: I've added a generic note now.
Also removed mentioning a term "archive" when talking about the file server's content, as using CPIO archive is just another implementation detail.

components/VM_Arm/src/main.c
```diff
@@ -426,9 +426,9 @@ static bool add_uts(const vm_config_t *vm_config, vka_t *vka, seL4_CPtr cap,
      *       ALLOCMAN_UT_DEV_MEM is, whether the UT can be requested from the
      *       allocator without providing it's backing physical address. Since
      *       UTs from the guest RAM region are supposed to be requested using
-     *       the physical address anyway, there seem not reason why they can't
+     *       the physical address anyway, there seems not reason why they can't
```
kent-mcleod, line 429:
```suggestion
     *       the physical address anyway, there seems no reason why they can't
```

axel-h: Thanks, fixed

kent-mcleod (approved):

PR #94, opened 2023-05-02
Odroid-C2: Fix GIC DTB path and provide GIC to VM by default

axel-h:
Odroid-C2: 
- Fix GIC DTB path 
- provide GIC to VM

To run the example on  Odroid-C2, the GIC is needed. This basically does what is done on the other platform also. If we have use cases where the GIC should explicitly not be provided, then we should make this a config setting in the VMM.


axel-h:
Can we merge this? Seems the device tree has change some time in the past. We can't get things working without this change.

kent-mcleod (approved):

PR #95, opened 2023-05-03
Fix typo in config variable name

axel-h:
This has done nothing so far, as the wrong config variable is set
KernelMaxNumBootinfoUntypedCap -> KernelMaxNumBootinfoUntypedCap**s**


lsf37 (approved):
This looks to be a relatively common typo, searching over the other repos I find 8 instances in total, most of them copy/pasted from the same source. I'll do a more thorough check and put up PRs for them.

camkes_vm_settings.cmake
```diff
@@ -11,7 +11,7 @@ set(KernelArch x86 CACHE STRING "" FORCE)
 set(KernelPlatform pc99 CACHE STRING "" FORCE)
 set(KernelVTX ON CACHE BOOL "" FORCE)
 set(KernelRootCNodeSizeBits 20 CACHE STRING "" FORCE)
-set(KernelMaxNumBootinfoUntypedCap 100 CACHE STRING "")
+set(KernelMaxNumBootinfoUntypedCaps 100 CACHE STRING "")
```
kent-mcleod, line 14: If the config option has been mistyped, then the line should be removed so that it doesn't change existing behavior.

lsf37: Or set to the default, yes, good point. Almost all instances seem to have been copy/pasted from the xx_verified configs, some of which set both values, probably because they got merged from different things.

lsf37: The default is currently 230:
https://github.com/seL4/seL4/blob/07a8263a9d5ccaafca9f94be1d3539a0cd3f9127/config.cmake#L276-L281

Since these config files will be taken as examples by people, we should think a bit about if we want to show that you can set the value here or if we should remove it, because init code should be able to deal with pretty much any reasonable default that comes from kernel settings.

kent-mcleod: I'm not sure why the config exists anymore. It used to clip the end of untyped's so that the init task could assume a lower bound on the number of free slots in the initial CNode (which the capdl-loader needs), but this was more of an issue back when the device untypeds were manually described for each individual device region in the SOC's manual and it was common to get hundreds of device untyped caps.

lsf37: Ok, in that case, I'd also vote for just removing the line. For the verified configs in the seL4, I've made a PR that will keep the value constant if the default is changed, so the proofs don't break, but we don't really need that here.

axel-h:
remove the line completely.

lsf37 (approved):

PR #96, opened 2023-05-07
vm_arm: remove deprecated config options

axel-h:
Test with: https://github.com/seL4/camkes-vm-examples/pull/31


axel-h:
This produces some strange errors with x86 at the moment because of the dependency on https://github.com/seL4/camkes-vm-examples/pull/31, which needs to be rebased to things work again.

PR #97, opened 2023-05-15
Cross-connector: use dataport frame size instead of 4k default

hlyytine:


axel-h:
Could you do a re-push, so we get another CI run? Should pass now with the dependencies.

axel-h:
Seems changing the label is sufficient to trigger a CI re-run. So let's see ...

axel-h (approved):

kent-mcleod (approved):

hlyytine:
Sorry, forgot to mention here that this shouldn't have been merged since I did not introduce that `frame_size_bits` additional field. We have two options: either revert this or shall I create a new PR that uses new getter that got merged? I think the latter is better, because it is more "proper" and I can create PR within an hour or two.

kent-mcleod:
Sure, happy for you to make the new PR.

PR #101, opened 2023-05-25
libvirtio: fix folder name for zynqmp platform

axel-h:
The folder name must be `virtioarm` and not `virtio`. See the other platform folders. Looks like there is nothing in CI that covers this, as this causes a compile time failure.

@chrisguikema Since you are using this platform also, how come you did not run into that? Do you use a different library?


axel-h:
Can we merge this?

abrandnewusername (approved):

chrisguikema:
@abrandnewusername can this be merged?

axel-h:
I have not tracked the PR actively since then, but it should still be fine as there have been no other changes since then.

PR #103, opened 2023-05-29
Use fdt_appendprop_uint() from libsel4vmmplatsupport

hlyytine:


axel-h:
I thinks we should close this ans use https://github.com/seL4/camkes-vm/pull/107 instead.

PR #113, opened 2023-10-25
init: add hooks for hpet emulation

chrisguikema:
This allows VMs to emulate the HPET for higher-precision timing.

Test with: https://github.com/seL4/seL4_projects_libs/pull/117, https://github.com/seL4/camkes-vm-examples/pull/51


chrisguikema:
Test with: https://github.com/seL4/seL4_projects_libs/pull/117

chrisguikema:
The build only seems to be failing because the "Test with" isn't working. Did I mess something up?

lsf37:
> The build only seems to be failing because the "Test with" isn't working. Did I mess something up?

The "Test with" has to be in the PR description, not in a comment. I'll update.

components/Init/src/main.c
```diff
@@ -438,6 +440,7 @@ static device_notify_t *device_notify_list = NULL;
 void pit_timer_interrupt(void);
 void rtc_timer_interrupt(uint32_t);
 void serial_timer_interrupt(uint32_t);
+void hpet_timer_interrupt(uint32_t);
```
abrandnewusername, line 443: Maybe add an `#ifdef CONFIG_VMM_USE_HPET` here

abrandnewusername (commented):
Looks great to me. It will be nice to have a VM example that demonstrates/tests HPET emulation.

components/Init/src/main.c
chrisguikema: Sure

chrisguikema:
> Looks great to me. It will be nice to have a VM example that demonstrates/tests HPET emulation.

Do we even have a test application for a 64-bit guest? If not, I can make one, and have it use the HPET emulation.

chrisguikema:
For what its worth, I added a flag to the `minimal` application that enables 64-bit guests and emulates the HPET. I don't know how to add that flag to the gitlab test runner, but it would be a good thing to add to the CI! @lsf37

chrisguikema:
Alright, so y'all will have to manually test the HPET emulation.

CI testing doesn't work because 64-bit VMs have a different file name than the Runner is expecting: https://github.com/seL4/camkes-vm/actions/runs/6697978462/job/18199249921#step:4:1299

And I can't test 32-bit VMs because they don't have HPET support compiled in.

chrisguikema:
@abrandnewusername Have you been able to look at this?

chrisguikema:
I updated the 32-bit Linux image to support the HPET, and it looks like the tests are failing. The `minimal` example does seem to use the HPET at least, so there is a chance that I messed something up.

`  vm_hpet_mmio_read@hpet.c:426 invalid HPET_CFG + 4`

@abrandnewusername do you want me to update the `UseHPET` variable to only be used with 64-bit guests? That way we can get this merged, and 32-bit support can be added later if anyone needs it?

chrisguikema:
@abrandnewusername I was able to work with @lsf37 to create a 64-bit guest test. The minimal_64 test worked on a hardware run, with the clocksource switching to the hpet: https://github.com/seL4/camkes-vm/actions/runs/6869868181/job/18683754338?pr=113#step:4:2791

What needs to happen to get this feature merged? Do I need to update the dependency to be a 64-bit guest?

components/Init/src/hpet.c
```diff
@@ -0,0 +1,739 @@
...
+ *
+ * *****************************************************************
+ *
+ * This driver attempts to emulate an HPET device in software.
+ */
+
+/*
+ * Copyright 2019, DornerWorks
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
```
lsf37, line 30: This should say `LGPL-2.0-or-later`

lsf37: (You''l then have to add the corresponding license text to `LICENSES/` for the license check to succeed -- you can use the attached [LGPL-2.0-or-later.txt](https://github.com/seL4/camkes-vm/files/13357934/LGPL-2.0-or-later.txt))

chrisguikema: Done!

abrandnewusername (approved):
I don't have much experience with HPET so I can't comment on the actual HPET emulation code. Apart from that, everything else is good to go.

PR #114, opened 2023-10-25
init, trivial: allow for relocs to be optional

chrisguikema:
Kernel relocations are not needed with 64-bit guest kernels. This commit allows for vm0.kernel_relocs = "" to specify no relocation is needed.


abrandnewusername (approved):

Indanz (approved):
I have no camkes experience, so I'm not to best person to review camkes change, but this seems okay. To me setting it to `strlen() != 0` would make more sense.

chrisguikema:
> I have no camkes experience, so I'm not to best person to review camkes change, but this seems okay. To me setting it to `strlen() != 0` would make more sense.

You could do that, but this was just copied from the `have_initrd` check here: https://github.com/seL4/camkes-vm/blob/master/components/Init/src/main.c#L683

chrisguikema:
Can this be merged?

PR #120, opened 2024-01-24
generic CI test

axel-h:
Test with: https://github.com/seL4/camkes-vm-images/pull/7, https://github.com/seL4/camkes-vm-examples/pull/57

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@
 name: Camkes VM Examples
 
 on:
-  pull_request_target:
+  pull_request_target: # !!!!!
     types: [opened, reopened, synchronize, labeled]
 
 # downgrade permissions to read-only as you would have in a standard PR action
--- a/components/VM_Arm/CMakeLists.txt
+++ b/components/VM_Arm/CMakeLists.txt
@@ -42,7 +42,7 @@ config_option(
     VM_VIRTIO_NET_ARPING
     "Enable virtio net arping module"
     DEPENDS
-    "KernelPlatformExynos5410 OR KernelPlatformExynos5422 OR KernelPlatformTx2 OR KernelPlatformQEMUArmVirt;VmPCISupport"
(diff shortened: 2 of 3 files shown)


axel-h:
@chrisguikema It seems to run the new test now, but there is an problem with the image name. Seems "zynqmp" is deep in the defaults. Either we clean this up to have a nice strong SOC/board differentiation or accept this as legacy for now to get things merged and have the board covered at least.

chrisguikema:
Axel, thanks for doing this! I think we're gonna need @lsf37's opinion here. When you build for the ultra96, or the zcu102, or any variant of the ZUS+ platform, the resulting image is always `capdl-loader-image-arm-zynqmp`. However, it looks like the CI scripts expect the `capdl-loader-image-arm-<SPECIFIC_PLATFORM>` instead. 

https://github.com/seL4/ci-actions/blob/master/camkes-vm/build.py#L59

We may be able to add another piece of logic in the `get_platform` function here: https://github.com/seL4/ci-actions/blob/08c98dc30fac13105aec51124a38206ab572944c/seL4-platforms/platforms.py#L147

But like I said, I'll wait for Gerwin to respond here, since there are a few ways we can go about this.

chrisguikema:
@axel-h can you rerun this? I added the `image_platform` which seems to exist solely to solve the problem we ran into.

axel-h:
just push another dummy commit to my branch https://github.com/axel-h/camkes-vm/tree/patch-axel-ci-1 then this PR's CI runs again.

chrisguikema:
I tried that, and it wouldn't let me push to your branch

axel-h:
ok, i will do a dummy push then latwr. You could make a new draft PR from your own branch that tragets https://github.com/seL4/camkes-vm/tree/patch-axel-ci-base instead of master, that will also work.

axel-h:
or well, removing and adding the `hw-test` label also triggered a re-run now

lsf37:
The CI support is merged now.

